PR #1, opened 2015-11-30
fix mbap payload length

timelesshaze:
Even though it's not copied to the payload, data already includes the CRC. This leads to an invalid payload length in the MBAP header.


yaacov:
Thanks :-)

PR #3, opened 2016-01-21
FC6 Preset Single Register

meteocontrol:
Hi Yaacov,
nice work with this module :)

We use it for a logger device prototype which heavily uses modbus. Therefore we needed the FC6 function to be implemented. I'am sorry that my IDE removed whitespaces so the merge request looks heavier than it actually is...
When you add `?w=false` to the url github will ignore the whitespace changes. e.g. https://github.com/yaacov/node-modbus-serial/compare/master...meteocontrol:master?w=false   

I hope you like the merge request.

Best regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# webstorm ide
+
+.idea/
+
 # Logs
 logs
 *.log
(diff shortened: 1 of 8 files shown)


yaacov:
Thanks for the FC6 support !
And for the spelling :-)

meteocontrol:
You're welcome!
Ty for merging so quickly ;-)

PR #4, opened 2016-01-26
Accomplished Modbus Function Code 16

Btotals:
Accomplished Modbus Function Code 15 "writeFC15", added promise method "writeCoils", also added and passed Mocha test case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -131,6 +131,16 @@ function _readFC6(data, next) {
         next(null, {"address": dataAddress, "value": value});
 }
 
+function _readFC15(data, next) {
+    var dataAddress = data.readUInt16BE(2);
+    var value = data.readUInt16BE(4);
+
+    if(next)
+        next(null, {"address": dataAddress, "value": value});
+}
+
+
+
 /**
  * Parse the data for a Modbus -
  * Preset Multiple Registers (FC=16)
@@ -272,6 +282,12 @@ ModbusRTU.prototype.open = function (callback) {
                  */
                 if (code == 6) {
                     _readFC6(data, next);
+                }
+
+                /* Force Multiple Coils
+                */
+                if (code == 15) {
+                    _readFC15(data, next);
                 }
 
                 // Preset Multiple Registers (FC=16)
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -131,6 +131,16 @@ function _readFC6(data, next) {
         next(null, {"address": dataAddress, "value": value});
 }
 
+function _readFC15(data, next) {
+    var dataAddress = data.readUInt16BE(2);
+    var value = data.readUInt16BE(4);
+
+    if(next)
+        next(null, {"address": dataAddress, "value": value});
```
lateral-thinker: "value" should be named "length" as this data tells the master the number of coils that were written with this operation.

yaacov:
Thanks, It's great to have FC15. 
Can you add FC15 support in all the ports that need specific function support (you added it only in testport) ?

ports/testport.js
```diff
@@ -185,7 +185,27 @@ TestPort.prototype.write = function (buf) {
...
+
+        // build answer
+        buffer = new Buffer(8);
+        buffer.writeUInt16BE(address, 2);
+        buffer.writeUInt16BE(length, 4);
 
+        // write coils
+        var byteCounts = Math.ceil(length / 8);
+        for (var i = 0; i < byteCounts; i++) {
+            this._coils &= buf.readUInt8(7 + i);
```
yaacov: The coils addresses are missing here ?
this._coils &= buf.readUInt8(7 + i);

index.js
Btotals: Yes, here the "value" should be "length". It was my carelessness, thanks you for the correction :-)

ports/testport.js
Btotals: Yes, the coil address would miss if another function code 15 comes. I care little at here. I would fix it, thanks for your correction :-)

Btotals:
I'm glad to contribute, is that means I should add _cmd 15 switch case in c701/rtubuffer/telnet port?

Btotals:
Fixed some bugs and updated README file.

PS: The origin README file has a little mistake at [readFC6](https://www.npmjs.com/package/modbus-serial#writefc6-unit-address-array-callback), the 3rd argument should be "value".

yaacov:
Great :-)

Btotals:
Hi yaacov:
Thanks you for this nice module, I was doing some work with Modbus protocol serval days ago and found your module on npm. And I found FC15 was extremely demanded in my work but this module couldn't supply, so I want to make a little contribution to this module.
I wish you would like my pull request and merge it.

Best Regards

yaacov:
I will merge it, full coverage of function codes is good thing to have :-) 
I do not have time now to look at it ... i will merge after I look at it.

Btotals:
Yes, full coverage is necessary, you've worked hard.

connium:
Hi Btotals,

in the README.md you missed to add FC 15 to the section "This class implements" and in the index.js there is no JSDoc for _readFC15. Can you please add the missing parts?

Unfortunately writeFC15 doesn't give me the results I expect. I try to figure it out what's going wrong here. So, stay tuned :)

connium:
Hi Btotals,

I had to rewrite the implementation of writeFC15 to figure out what was wrong. The following implementation works fine.

    var dataBytes = Math.ceil(array.length / 8);
    var codeLength = 7 + dataBytes;
    var buf = new Buffer(codeLength + 2);  // add 2 crc bytes
    
    buf.writeUInt8(address, 0);
    buf.writeUInt8(code, 1);
    buf.writeUInt16BE(dataAddress, 2);
    buf.writeUInt16BE(array.length, 4);
    buf.writeUInt8(dataBytes, 6);

    var bufferOffset = 7;
    var bitString = array.map(function(item) {
        return item ? '1' : '0';
    });

    while (bitString.length > 8) {
        buf.writeUInt8(parseInt(bitString.splice(0, 8).join(''), 2), bufferOffset++);
    }
    buf.writeUInt8(parseInt(bitString.join(''), 2), bufferOffset);

    // add crc bytes to buffer
    _CRC16(buf, codeLength);

    // write buffer to serial port
    this._port.write(buf);

This will convert the boolean array to a string array, parse the string by byte and write it to the buffer. Feel free to copy it.

Btotals:
@connium 
Hi connium,
although my implementation of FC15 works well in my project, I think your implementation is more elegent, thanks your for the correction. But could you tell me in which case my code would give unexpected result?

connium:
Hi @Btotals,

I tested with several values and always got unexpected data values.

One example:
Slave ID: 17, write coils #20 to #29 (1100 1101 01)
This should return an RTU message of 11 0F 0013 000A 02 CD01 BF0B, where CD01 is the coil values. I'm at my personal PC and cannot test any code right now, so I'm not able to give you the result of your implementation. At least it wasn't the message mentioned above - it always was the coil values that didn't match my expectations.

Btotals:
Hi, @connium 
Sorry, in your case here might has something wrong. If write coils from **#20**, start address should be 0014 rather than 0013. And here your case 1100 1101 01, which is works fine with my code, here is codepen [link](http://codepen.io/Btotals/pen/BjVpqG)

connium:
According to the Modbus specification the first data address is x0001 and not x0000 as one would expect in informatics. Therefore coil #1 is addressed as 0x00 and coil #20 corresponds to 0x13. This is valid for all function codes.

For me it doesn't feel intuitive to reverse the input array in order to get the expected result. Eventually I'm wrong.

yaacov:
Thanks @Btotals and @connium , pull request merged with @connium suggestions. 
I did not understand the addressing issue you talked about, hope to look at it soon.

PR #12, opened 2016-02-11
Add a check to see if the configured port is open before attempting a…

dom-white:
To avoid any possibility of attempting a write on a non-open port I have added a test to check the port has been opened successfully.
With the serial ports there is a simple isOpen() method which I have tied into. With the other ports I have added a flag which is set/unset by the addition of .on('opened') and .on('closed') signals provided by the udp or tcp protocols.
Hope this makes sense.

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,6 +295,13 @@ ModbusRTU.prototype.writeFC2 = function (address, dataAddress, length, next, cod
     // function code defaults to 2
     code = code || 2;
 
+    // check port is actually open before attempting write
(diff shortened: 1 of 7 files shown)


review-ninja:
<a href="https://app.review.ninja/yaacov/node-modbus-serial/pull/12" target="_blank">ReviewNinja

connium:
Like :thumbsup:

yaacov:
Thanks :+1:

PR #15, opened 2016-02-15
fix isOpen method where extra flag is required to set connection state

dom-white:
I seem to have broken the ethernet based ports (telnet,tcp,c701) when I added the isOpen() method in.
When testing they respond as closed even when open as I had set the flag without the correct context pointer.
This has now been fixed, sorry for the trouble.

### DIFF
--- a/ports/c701port.js
+++ b/ports/c701port.js
@@ -75,11 +75,11 @@ var UdpPort = function(ip, options) {
     });
 
     this._client.on('listening', function() {
-        this.openFlag = true;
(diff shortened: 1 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/yaacov/node-modbus-serial/pull/15" target="_blank">ReviewNinja

yaacov:
it was a .0 version, they are always buggy. now it is .1

dom-white:
very understanding of you :-)

PR #18, opened 2016-02-22
Fix exception handling for buffered ports and add tests for it.

connium:
The buffered ports (RTU, Telnet) don't pass Modbus exceptions to the Modbus class, because they only accept responses with the expected length. This PR will fix this (added tests for it).

I removed the CRC checking from both ports. It is done in the main class anyway and there will raise an error while the buffered ports would eat up the message silently.

### DIFF
--- a/ports/rtubufferedport.js
+++ b/ports/rtubufferedport.js
@@ -3,34 +3,13 @@ var util = require('util');
 var events = require('events');
 var SerialPort = require("serialport").SerialPort;
 
-var crc16 = require('./../utils/crc16');
-
-/**
- * check if a buffer chunk can be a modbus answer
- * or modbus exception
- *
- * @param {RTUBufferedPort} modbus
- * @param {Buffer} buf the buffer to check.
- * @return {boolean} if the buffer can be an answer
- */
-function checkData(modbus, buf) {
-    // check buffer size
-    if (buf.length != modbus._length && buf.length != 5) return false;
-
-    // calculate crc16
-    var crcIn = buf.readUInt16LE(buf.length - 2);
-
-    // check buffer unit-id, command and crc
-    return (buf[0] == modbus._id &&
-        (0x7f & buf[1]) == modbus._cmd &&
-        crcIn == crc16(buf.slice(0, -2)));
-}
+var EXCEPTION_LENGTH = 5;
 
 /**
  * Simulate a modbus-RTU port using buffered serial connection
  */
 var RTUBufferedPort = function(path, options) {
-    var modbus = this;
+    var self = this;
 
     // options
     if (typeof(options) == 'undefined') options = {};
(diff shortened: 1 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/yaacov/node-modbus-serial/pull/18" target="_blank">ReviewNinja

yaacov:
Thanks, It was ugly code.

dom-white:
I am finding an issue with this code change. I can no longer get data reliably over the rtubuffered port from a modbus slave.

The issue appears (for me) when we have a larger payload response followed by a smaller payload response.

In my case I send off a request with an expected response payload of length 9 and I get back the following buffer (in the _emitData  i=0 length = 9)
<Buffer 01 03 04 00 5a 07 d0 d9 8c>
I then send a request with an expected payload of length 6. This get ignored somehow, and when I send a new request with an expected payload of 9 again the whole buffer looks like this
<Buffer 01 02 01 55 61 b7 01 03 04 00 5a 07 d0 d9 8c>
This has my ignored 6 bytes at the start and then the data from the following 9 bytes response
The index given to _emitData in this case is now 6 and the length is 9, so it extracts the 9 byte response correctly from the later part of the buffer. It is just that it has missed the previous 6 byte response that I was expecting first. 
i.e. <Buffer 01 02 01 55 61 b7 >

I am not sure why this is the case yet, I have not had time to have a proper look at it, I am just rolling back the version I am using for now

yaacov:
@dom-white Thanks :+1: , I will look into it too.

connium:
Hm, I didn't run into such behavior. But will also dig into it on monday. Thanks for reporting this.

@dom-white can you provide the unit IDs and the function codes you are working with? (just in case this causes the code to break)

yaacov:
Hypothetically one can call the write function after `this.emit('data', ...)` and before `this._id = 0;`.
@dom-white , @connium can you check https://github.com/yaacov/node-modbus-serial/commit/da1b3c82e3eff4b5894dc49f487cbafcd18fb1d8 to see if it has effect on the problem ?

dom-white:
I will have a further look on Monday, maybe it is just me, or the data I am passing it.

Thanks,
Dom

dom-white:
Well spotted, this is exactly what is happening for me. I have a chain of writes I need to do, and I trigger the next one as soon as I get a successful response data back from the previous request. 

Adding more instrumentation to the code, I clearly see the 'write' function setting these variables, then the post emit code resetting them to zero.
So you latest check in fixes the issue for me.
Thanks,
Dom

yaacov:
:+1:
fixed by https://github.com/yaacov/node-modbus-serial/commit/da1b3c82e3eff4b5894dc49f487cbafcd18fb1d8
**closed**

PR #20, opened 2016-03-01
Make EventEmitter usage compatible with legacy NodeJS (0.10.x, 0.12.x)

relovution:
Without this, you'll see this cryptic error:

ctor.prototype = Object.create(superCtor.prototype, {
                          ^
TypeError: Object prototype may only be an Object or null

Any time one requires this module. What changed after 0.12.x is that whereas before EventEmitter was a class within the event module, it has now been moved to the top level.

### DIFF
--- a/ports/asciiport.js
+++ b/ports/asciiport.js
@@ -1,6 +1,7 @@
 'use strict';
 var util = require('util');
 var events = require('events');
+var EventEmitter = events.EventEmitter || events;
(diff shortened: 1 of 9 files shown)


review-ninja:
<a href="https://app.review.ninja/yaacov/node-modbus-serial/pull/20" target="_blank">ReviewNinja

yaacov:
Thanks, thats a nice touch :+1:

PR #35, opened 2016-07-28
Update to be compatible with v4 branch of node-serialport

dom-white:
As per the upgrade guide
https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/4.0.1/UPGRADE_GUIDE.md

1 deprecation and 1 breaking change (if running with autoconnect disabled as modbus is).

I have tested this with rtubuffered mode, and applied the change to rtu and ascii modes too

### DIFF
--- a/apis/connection.js
+++ b/apis/connection.js
@@ -37,10 +37,12 @@ var addConnctionAPI = function(Modbus) {
             next = options;
             options = {};
         }
+        // disable auto open, as we handle the open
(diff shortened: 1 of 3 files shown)


yaacov:
Thanks

yaacov:
v4.5.0 is out 
Thanks Dominic :+1:

dom-white:
cool, thanks for the super fast turn around

PR #47, opened 2017-01-29
return error objects

yaacov:
Return error objects on error instead of strings.

We currently return strings to indicate error, this is inconsistent with the tcp and serial open error messages, this will make the errors we return, Error objects.    

@connium this is a breaking change ( if someone check for the string error value ) please review.

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,7 +144,7 @@ function _startTimeout(duration, next) {
     }
     return setTimeout(function () {
         if (next) {
-            next('Timed out');
+            next(new Error('Timed out'));
         }
     }, duration);
 }
@@ -220,7 +220,7 @@ ModbusRTU.prototype.open = function (callback) {
                     error = "Data length error, expected " +
                         transaction.nextLength + " got " + data.length;
                     if (transaction.next)
-                        transaction.next(error);
+                        transaction.next(new Error(error));
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -401,8 +401,10 @@ ModbusRTU.prototype.writeFC4 = function (address, dataAddress, length, next, cod
 
     // check port is actually open before attempting write
     if (this._port.isOpen() === false) {
-        var error = "Port Not Open";
-        if (next) next(error);
+        if (next) {
+            var message = "Port Not Open";
+            next(new Error(message));
```
connium: The `message` variable is not needed here. IMO it's more readable if you write

```js
next(new Error("Port not open"));
```

index.js
```diff
@@ -442,8 +444,10 @@ ModbusRTU.prototype.writeFC5 = function (address, dataAddress, state, next) {
...
```
connium: The `message` variable is not needed here. IMO it's more readable if you write

```js
next(new Error("Port not open"));
```

index.js
```diff
@@ -488,8 +492,10 @@ ModbusRTU.prototype.writeFC6 = function (address, dataAddress, value, next) {
...
```
connium: The `message` variable is not needed here. IMO it's more readable if you write

```js
next(new Error("Port not open"));
```

index.js
```diff
@@ -586,8 +592,10 @@ ModbusRTU.prototype.writeFC16 = function (address, dataAddress, array, next) {
...
```
connium: The `message` variable is not needed here. IMO it's more readable if you write

```js
next(new Error("Port not open"));
```

Issue #46, opened 2017-01-22
Modbus request in response


biancode:
A [feature request](https://github.com/biancode/node-red-contrib-modbus/issues/3) needs an enhancement of the response to have data, response and request buffer, please

response {
 request: Buffer with CRC
 data : []
 buffer: Buffer
}

That is a good idea for debugging

yaacov:
hi,

can you try to take:
https://github.com/yaacov/node-modbus-serial/blob/master/ports/rtubufferedport.js
copy it call it debug port.

then add an emiter for `debug` messages , e.g.
when reciving data [line 72]
`this.emit('debug', this._buffer.slice(start, start + length));`
and when sending data [line 136]
`this.emit('debug', data);`

I am a little busy with other things, so if you will wait for me, it can take some time,
I think that this new debug port will do the trick :-)

yaacov:
fixed by #50 and #51

yaacov:
@connium @biancode hi,

Looking at the  [debug](https://www.npmjs.com/package/debug) module I only see the options to send to stdout/file.
Can we solve this issue using  [debug](https://www.npmjs.com/package/debug) module or some other standard way ?

biancode:
I'll try some more next week.

biancode:
for me yes - debug is perfect - but there is no debug output for now - right?  DEBUG=*  or DEBUG=modbus-serial doesn't work

connium:
You are right, there are no debug outputs in the code yet. Will add them in the upcoming week (I'm missing them sometimes as well :wink: ).

yaacov:
> I'm missing them sometimes as well 😉

👍

biancode:
node-red-contrib-modbus is working with debug npm from version 1.0.6+ and using the name 
`DEBUG=modbus-serial `
for modbus info and otherwise with the name 
`DEBUG=node_red_contrib_modbus`

PR #50, opened 2017-01-29
Add optional debug data on send and recive data

yaacov:
Add optional debug data on send and recive data

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,14 @@ Duration of the timeout
 
 
 ----
+##### .setDebug(callback)
+      Set the port to emit debug messages into callback
+
+*callback {function}:*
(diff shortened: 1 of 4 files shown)


connium:
Instead of adding a `debug` callback I would prefer to use the well-known node module [debug](https://www.npmjs.com/package/debug).

This is used by many node modules and allows you to trace the module you are interested in or to debug your entire application.

yaacov:
@connium @biancode hi,

In https://github.com/yaacov/node-modbus-serial/issues/46 @biancode wants to use the debug strings for use in another function.

Looking at the debug module I only see the options to send to stdout/file.

Can we solve https://github.com/yaacov/node-modbus-serial/issues/46 using debug module or some other standard way ?

PR #51, opened 2017-01-29
Add an example of debug usage

yaacov:
Add an example of debug usage


examples/debug.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // run program
+    run();
+}
+
+function run() {
+  // read the 4 registers starting at address 5
+  client.readHoldingRegisters(5, 4)
+    .then(function(d) {
+      console.log("Recive:", d.data); })
```
connium: typo: `Receive`

Issue #44, opened 2017-01-13
Is it possible to enhance this package with slave/server


biancode:
Is it possible to enhance this package with slave/server, please?

http://www.rtaautomation.com/technologies/modbus-tcpip/

yaacov:
> Is it possible to enhance this package with slave/server

a. yes :-)
b. what is the use case ?
c here is mock server, is this what you have in mind ?
```
// create an empty modbus server
var ModbusRTU = require("modbus-serial");
var server = new ModbusRTU();

// open connection to a tcp line
// and register some response functions 
client.serveTCP({url: "0.0.0.0", port: 502} , { 
    readInputRegisters: function(address, length) { return [42; ]; },
    writeCoil: function(address, state) { console.log("writing coil"); } 
});

```

biancode:
a. great!
b. https://github.com/biancode/node-red-contrib-modbus/blob/master/src/modbus-server.js
want to use only modbus-serial in node-red package
c. looks good and also see https://github.com/biancode/node-red-contrib-modbus/blob/master/src/modbus-server.js from line 111

yaacov:
fixed by #52

see example:
https://github.com/yaacov/node-modbus-serial/blob/master/examples/server.js

biancode:
cool - I'll start testing next week

PR #52, opened 2017-01-29
Add server support

yaacov:
**Add server support**

Add an option to run a Modbus server.
issue: https://github.com/yaacov/node-modbus-serial/issues/44

### DIFF
--- a/servers/servertcp.js
+++ b/servers/servertcp.js
@@ -1,0 +1,269 @@
+'use strict';
+/**
+ * Copyright (c) 2017, Yaacov Zamir <kobi.zamir@gmail.com>
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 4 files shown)


servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+ var EventEmitter = events.EventEmitter || events;
+ var net = require('net');
+
+ var HOST = '127.0.0.1';
+ var MODBUS_PORT = 502;
+
+/* Add bit operation functions to Buffer
+ */
+require('../utils/buffer_bit')();
+var crc16 = require('./../utils/crc16');
```
connium: `require('../utils/crc16');`

servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Parse a modbusRTU buffer and return an answer buffer
+ */
+function parseModbusBuffer(buf, vector) {
```
connium: please use speaking variable names:
`requestBuffer` instead of `buf`

servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+    var ansBuf = null;
```
connium: `responseBuffer`

servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+    var unitNumber = buf[0];
```
connium: Most of the time you use `unitID` in the existing code. I guess it makes sense to stay consistent in the naming of things.

servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+        var length = buf.readUInt16BE(4);
+
+        // if length is bad, ignore message
+        if (buf.length != 8) {
+            return;
+        }
+
+        // build answer
+        ansBuf = new Buffer(3 + parseInt((length - 1) / 8 + 1) + 2);
+        ansBuf.writeUInt8(parseInt((length - 1) / 8 + 1), 2);
```
connium: You use the same calculation twice. It's more readable if you extract it to a local variable:
`var dataLength = parseInt((length - 1) / 8 + 1)`

servers/servertcp.js
```diff
@@ -0,0 +1,269 @@
...
+
+        // read coils
+        if (vector.getCoil) {
+            for (var i = 0; i < length; i++) {
+              ansBuf.writeBit(vector.getCoil(address + i), i, 3);
```
connium: This will work only if 1-8 coils are requested as you always write to the 4th byte of the response buffer.

PR #60, opened 2017-02-10
Corrected modbus frame header detection

SuperJojo2001:
Code now stops searching for valid frame header if one found and then waits for expected frame length

### DIFF
--- a/ports/rtubufferedport.js
+++ b/ports/rtubufferedport.js
@@ -46,17 +46,31 @@ var RTUBufferedPort = function(path, options) {
             var unitId = self._buffer[i];
             var functionCode = self._buffer[i+1];
 
-            if (unitId !== self._id) continue;
-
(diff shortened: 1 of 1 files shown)


ports/rtubufferedport.js
```diff
@@ -46,17 +46,31 @@ var RTUBufferedPort = function(path, options) {
...
-
-            if (functionCode === self._cmd && i + expectedLength <= bufferLength) {
-                self._emitData(i, expectedLength);
-                return;
-            }
-            if (functionCode === (0x80 | self._cmd) && i + EXCEPTION_LENGTH <= bufferLength) {
-                self._emitData(i, EXCEPTION_LENGTH);
-                return;
+            if ( unitId === self._id) {
+                if( functionCode === self._cmd || functionCode === (0x80 | self._cmd)) {
```
yaacov: `if (a) if (b)` pattern is not nice ...
can you make it `if ( (a) && (b) )` ?

PR #65, opened 2017-02-23
Enable Modbus TCP Server to support multiple unit id's

mwillbanks:
Presently the Modbus TCP Server only provides the option of a unit id in
the constructor.  This option is never utilized nor enforced and is
misleading.  The unit id from the request shall be passed to the *vector*
methods to allow for responses for various units on the same IP address
and port.

The option for the constructor is dropped in this case as it is defunct.


yaacov:
👍 thanks !

PR #66, opened 2017-03-03
tcp buffered and improved telnet

biancode:
please, test this very hard - I tested with the PLC Simulator for Windows and the test cases, but for now I have no hardware to test - next week I'll get a new one, so maybe you can give me some response, how it works for you, please

### DIFF
--- a/apis/connection.js
+++ b/apis/connection.js
@@ -20,34 +20,34 @@
  *
  * @param {ModbusRTU} Modbus the ModbusRTU object.
  */
-var addConnctionAPI = function(Modbus) {
+var addConnctionAPI = function (Modbus) {
 
     var cl = Modbus.prototype;
 
-    var open = function(obj, next) {
-      /* the function check for a callback
-       * if we have a callback, use it
-       * o/w build a promise.
-       */
-      if (next) {
-          // if we have a callback, use the callback
-          obj.open(next);
-      } else {
-          // o/w use  a promise
-          var promise = new Promise( function (resolve, reject) {
-              function cb(err) {
-                  if (err) {
-                      reject(err);
-                  } else {
-                      resolve();
-                  }
-              }
+    var open = function (obj, next) {
+        /* the function check for a callback
+         * if we have a callback, use it
+         * o/w build a promise.
+         */
+        if (next) {
+            // if we have a callback, use the callback
+            obj.open(next);
+        } else {
+            // o/w use  a promise
+            var promise = new Promise(function (resolve, reject) {
+                function cb(err) {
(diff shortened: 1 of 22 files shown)


apis/connection.js
```diff
@@ -20,34 +20,34 @@
  *
  * @param {ModbusRTU} Modbus the ModbusRTU object.
  */
-var addConnctionAPI = function(Modbus) {
+var addConnctionAPI = function (Modbus) {
```
yaacov: hi,

can yo clean up the changes that are not really changes ?
like this extra space, and the extra spaces in line 27..56, they make it harder to understand what is going on here :-)

yaacov:
hi, great !!!
the debug is grate too 👍

yaacov:
please take the time to remove changes in the file that are not part of this pull request code ( e.g. extra spaces and variable name changes ) 
this changes make it really hard to understand what is new code, and what is just irrelevant changes.

servers/servertcp.js
```diff
@@ -234,7 +235,8 @@ var ServerTCP = function (vector, options) {
             requestBuffer.writeUInt16LE(crc, requestBuffer.length - 2);
 
             // emit debug data
-            if (modbus.debug) modbus.emit('debug', {action: 'recive', data: requestBuffer});
+            modbusSerialDebug(JSON.stringify({action: 'receive', data: requestBuffer}));
```
yaacov: @biancode hi, now that we use the `debug` module, can we remove the `modbus.emit('debug') ...` ?
if using `debug` fix https://github.com/yaacov/node-modbus-serial/issues/46 then we should drop the 'debug' emits ?

biancode:
Please give me your code formatter settings

yaacov:
> Please give me your code formatter settings

👍 i do it without formatter :-) , sometimes i also do mistakes because of it, but this is what I'm used to ...

My requests
1. Be consistent with the rest of the code, fix your formatter until it doesn't change the current code.
2. Even if the code you see is wrong ( not consistent with most of the code ) don't fix it in this pull request, we can fix it in another pull request that will only fix lintting.

yaacov:
@biancode this is a very good point, format should be automated ! opend this issue https://github.com/yaacov/node-modbus-serial/issues/67

servers/servertcp.js
biancode: Do we need it to be backward compatible? If not, then yes we should delete it.

yaacov:
@biancode do you need help re-base your code after https://github.com/yaacov/node-modbus-serial/pull/68 , to work out the conflicts ?

PR #78, opened 2017-03-04
consistent README and package text

biancode:


yaacov:
:-)

biancode:
ah saw it to late - but that's it for now :)

PR #89, opened 2017-03-15
move serialport from devDependencies to dependencies

### LINKED ISSUE #88: `serialport` listed in devDependencies
[here's a direct link.](https://github.com/yaacov/node-modbus-serial/blob/52390e87d7b9ee5cb354abafb48acff62f5b7bc1/package.json#L37)

since it's not in the `dependencies` list, when i `npm install modbus-serial`, `serialport` doesn't get installed, leaving this package in a broken state.

### solution

move `serialport` into `dependencies`

thanks!! 🌟

brandly:
fixes #88. i should've just sent this PR in the first place 

thanks a lot for this lib!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,10 @@
     "mocha-eslint": "^3.0.1",
     "mockery": "^2.0.0",
     "pump": "^1.0.2",
-    "serialport": "4.0.7",
     "sinon": "^1.17.7"
   },
   "dependencies": {
-    "debug": "^2.6.1"
+    "debug": "^2.6.1",
+    "serialport": "4.0.7"
   }
 }


yaacov:
Thanks 👍 

We had a discussion about it, I didn't want any dependencies, but now we have `debug` as a dependency, and I do not think it's that bad to drag `serialport` along too.

I will merge this, we can always revert :-) in case this dependency is problematic in other use cases (when you only use TCP/UDP ports)

cc @connium @biancode

yaacov:
p.s.

@brandly we also need to: 
a. Update the README to not [tell people](https://github.com/yaacov/node-modbus-serial/blame/master/README.md#L33) [1] to install `serialport`separately.
b. remove the [`try ... catch`](https://github.com/yaacov/node-modbus-serial/blob/master/index.js#L632) [2] we have in case people do not want to install `serialport`, now we do it for them :-)

can you do that in a new pull request ?

[1] https://github.com/yaacov/node-modbus-serial/blame/master/README.md#L33
[2] https://github.com/yaacov/node-modbus-serial/blob/master/index.js#L632

PR #100, opened 2017-05-05
the serial ascii connect method should return a promise

ilang:
like all other connects the ascii connect should return a promise.
(Actually Node Red uses this prmise, and hence fails to open ascii
serial connection)


yaacov:
Wooo, tanks, how we missed that 👍

yaacov:
Published v5.2.10 with your fix to npm, thanks @ilang 🍰 

cc//  @biancode ( this affect nod-red module )

biancode:
done v1.0.12

PR #103, opened 2017-05-07
start testing server

biancode:
it is not complete but a good point to start

### DIFF
--- a/ports/tcpport.js
+++ b/ports/tcpport.js
@@ -119,10 +119,8 @@ TcpPort.prototype.write = function(data) {
     var transactionsId = (this._transactionId + 1) % MAX_TRANSACTIONS;
 
     // remember current unit and command
-    this._id = data[0];
-    this._cmd = data[1];
-
-    // TODO: is it not interesting to have the expected length here as in all other ports?
+    var unitNumber = data[0];
+    var functionCode = data[1];
 
     // remove crc and add mbap
     var buffer = new Buffer(data.length + MIN_MBAP_LENGTH - CRC_LENGTH);
@@ -138,8 +136,8 @@ TcpPort.prototype.write = function(data) {
         action: "send tcp port",
         data: data,
         buffer: buffer,
-        unitid: this._id,
-        functionCode: this._cmd,
+        unitid: unitNumber,
+        functionCode: functionCode,
         transactionsId: transactionsId
     });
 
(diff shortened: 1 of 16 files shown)


biancode:
The test of server will fail, but this is for you to check please, if I am wrong in that case or if there is a bug.

biancode:
The test of the server will fail, but this is for you to check please, if I am wrong in that case or if there is a bug.

yaacov:
Thanks 👍 

I will try to look at it today ... If I will not have time today, I will look at it next weekend.

biancode:
take your time - it is not mission critical

ports/tcpport.js
```diff
@@ -118,6 +118,12 @@ TcpPort.prototype.write = function(data) {
     // get next transaction id
     var transactionsId = (this._transactionId + 1) % MAX_TRANSACTIONS;
 
+    // remember current unit and command
+    this._id = data[0];
+    this._cmd = data[1];
+
+    // TODO: is it not interesting to have the expected length here as in all other ports?
```
yaacov: A. In the case of modbus tcp, we are not guaranteed a `request -> response` order, like other ports.

For example we can have a `request_12 `and then get `response_1` and then `response_4` and then `request_13` ... 
We can not trust the values in `this._id` , `this._cmd` and/or `this._length` to correspond to the next response.

B. We save the `transactionId` , so we can store this values for this transaction, and retrieve the appropriate values once we read the `transactionId` of the response packet.

C. I think it's not safe to store `this._id` and `this._cmd` here, because the next response may not be the response for this request. We do not test for this sequence of events ... so the tests will not catch that, but it's wrong.

yaacov:
Hey, some commits here I can merge without looking too much at the code ( like the doc strings ) and some I need to read and think about ...

If you make them in separate pull requests I can merge the ones that are simple and wait only with the ones I need to read more carefully.

biancode:
Next time I'll use feature branches to send pull requests one by one.

yaacov:
a. Assigned myself so I will not forget ...

@biancode can you re-open this pull reqest as a series of smaller pull requests by topic, I do not find the time to understand this one big pull reqest ... this way I can merge the small things, and maybe the big one will not look so scary :-)

biancode:
Yes, but it will take some time.

yaacov:
> Yes, but it will take some time.

Thanks, currently it's just too big for one issue :-)

biancode:
I tried GitKraken and it is very cool to work with Git Flow and Feature Branches, so I'll do it this way based on the actual state of your repo.

PR #108, opened 2017-05-25
server and server testing

biancode:
started server testing, which should work but doesn't - given tests should pass


biancode:
remember test will fail but it is for you to work on these tests please

yaacov:
Thanks 👍 , I will fix the tests in a new PR

PR #136, opened 2017-09-09
Vmin improvement

ReinierH:
added and tested vmin property to Serialport to prevent Serialport from emitting events when receiving bytes < MIN_MODBUS_FRAME.


ReinierH:
definetely time to add eslint to vim :)

yaacov:
Thanks !

p.s.
next pull request, if you have time, make it check if user didn't already set this options to a different value :-)

p.p.s
if you need me to make a new npm package with this change, ping me 👍

Issue #137, opened 2017-09-23
server need working close


biancode:
Hi,

the modbus._server doesn't close and so I get:

> Error: listen EADDRINUSE 0.0.0.0:11502

idea after a short search...

```
ModbusRTU.prototype.close = function(callback) {
    // close the serial or net port if exist
    if (this._port) {
        this._port        this._port.close(callback);
        this._port.removeAllListeners("data");
    }
   // close the net port if exist
    if(this._server) {
        this._server.close(callback);
        this._server.removeAllListeners("data");
   };
```

It needs a [server.close()](https://nodejs.org/api/net.html#net_server_close_callback), please.

PR #140, opened 2017-09-24
add a close method to ServerTCP

yaacov:
fix: https://github.com/yaacov/node-modbus-serial/issues/137


yaacov:
@biancode please check if this solves #137

it can be used like this:

```
var serverTCP = new ModbusRTU.ServerTCP(vector, { host: "0.0.0.0", port: 8502, debug: true, unitID: 1 });

serverTCP.on("socketError", function(err) {
    console.error(err);
    serverTCP.close(someCallbackMethod)
});
```

PR #141, opened 2017-09-24
new server example with buffers

biancode:


yaacov:
hi :-)

did you see https://github.com/yaacov/node-modbus-serial/pull/140, if you approve it, I will merge and you can use it in this example

yaacov:
I prefer the more complete "logger" and "write" example to be in new files with names like ?"logger-complete"?  , this way we will have small / simple example files, **and** more complete examples showing more options.

biancode:
OK

Issue #142, opened 2017-09-24
new NPM release


biancode:
Dear yaacov,

Please, don't forget the NPM release. I need the server close for Node-RED.

yaacov:
fixed by #143

PR #143, opened 2017-09-24
bump minor version

yaacov:
backword compatible api changes:
-  add `close` method to server
-  add minimal frame size check in rtu buffer

bugfixes:
-  add example
-  bug fixes


yaacov:
fix: #142

Issue #145, opened 2017-09-25
server response FC3 and FC4 wrong


biancode:
Dear yaacov,

please, try the new buffered_server and the write_complete and logger_complete clients.
The server response for FC3 and FC4 is not correct in that cases.

cya
Klaus

biancode:
bufferFactor = 8 was missing

PR #146, opened 2017-09-25
fixed examples with buffer factor

biancode:
fix for #145


yaacov:
hi, thanks.

a. update major version - because of using new serial port version ? should be in a different pull request.
b. why support only LTS versions ?

biancode:
a) OK I'll split
b) LTS for platform support serialport see [README](https://www.npmjs.com/package/serialport)

Issue #149, opened 2017-09-30
serialport isOpen to Boolean - is not a function


biancode:
I did not see port.isOpen - [Serialport](https://github.com/EmergingTechnologyAdvisors/node-serialport/blob/master/test/serialport.js) line 219 test - is now Boolean not a function

biancode:
hint for me - Serialport should be wrapped to hold a static own API

biancode:
@yaacov what do you say - change isOpen to Boolean in that lib or build a wrapper for SerialPort to be compatible

yaacov:
> @yaacov what do you say - change isOpen to Boolean in that lib or build a wrapper for SerialPort to be compatible

I prefer to change isOpen to Boolean like:
https://github.com/EmergingTechnologyAdvisors/node-serialport/commit/5616664ed67a983a28aa54e08350f374e9be9085

yaacov:
@biancode hi,

We have a new version v6.0.1 that fixs this 🎉 🍰 
Can you check that it fixes this issue ?

biancode:
@yaacov - yes I'll check it tomorrow - thank you!

biancode:
@yaacov it works very well for me and for now, I'm very happy with that version

If you find some time try your Modbus server in Node-RED

![image](https://user-images.githubusercontent.com/4369867/31099604-dd91ac00-a7c6-11e7-87d6-af6def70b82b.png)

yaacov:
> it works very well for me

PR #150, opened 2017-10-01
Make isOpen property

yaacov:
**Description**

As of v5.0.0 SerialPort changed isOpen to be property instead of function.
We need to update ModbusSerial to work with the new version of SerialPort.


yaacov:
fix #149

PR #156, opened 2017-10-13
Using backword incompatible serial port, bumping major version

yaacov:
Since #155 we are using a none backward compatible serialport version.

No change to api, but this is none backward compatible change for users who relay on changed serialport functionality.


yaacov:
cc @reconbot

PR #157, opened 2017-10-20
Modbus Enron

LeonardoPilato:
Is possible modify this library for use with Modbus Enron/Daniel? Is something hard to do? I try use with a device with Modbus RTU Enron but I get errors on data length or in CRC validation. Also Enron Modbus has Historic data for Hour and Day averages is possible create a function for poll this historic data? Can you help me with this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ This class makes ModbusRTU (and TCP) calls fun and easy.
 Modbus is a serial communications protocol, first used in 1979.
 Modbus is simple and robust, openly published, royalty-free and
 easy to deploy and maintain.
+--- 4.5.3 automatically remove 00 at the beginning
+--- 4.5.2 Add modbus rtu over tcp: use connectComOverTCP("ip:port",{},function(error){xxx}}  to connect tcp port
+    in this way, you can use serial-ethernet connector
+    It will reconnect to server at 10s interval on broken.
+
 
 ----
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # modbus-serial master
 A pure JavaScript implemetation of MODBUS-RTU (and TCP) master for NodeJS
+!!! this is a forked version !!!
+in the original version, the parser only check and parse one piece of data, which means an error will occur if the data is emitted more than once.
+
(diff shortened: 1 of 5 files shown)


yaacov:
> Is possible modify this library for use with Modbus Enron/Daniel?
 > is possible create a function for poll this historic data?

The Enron commands are the same as Modicon commands, with different meaning to the 
addresses, and data length.

For example: a history request will issue a standard FC3 Modicon request, with a length of 701
registers, but the controller will respond with a standard FC3 responce that has only 10 registers.

> Is something hard to do?

It may require changes to the way we check the response length, because Enron responses may
have different length then Modicon responses for the same request.

You can start by removing the code [1] that check the responses length and address, and see 
what happen.

> Can you help me with this?

No, I do not have an Enron controller or Time :-)

[1] https://github.com/yaacov/node-modbus-serial/blob/master/index.js#L275

yaacov:
The code in this pull request has conflicts that must be resolved

@LeonardoPilato , Please fix the code conflicts, code errors and styling.

yaacov:
@LeonardoPilato 

a. This pull request is about adding a new port for RTU over TCP.
a.1 The title of this pull request is not what this pull request is about.
a.2 We already have have the telnet [1] port for that, what does this new port add over the telnet port ?

b. This pull request include changes to the README and package files that are not required here.

[1] https://github.com/yaacov/node-modbus-serial/blob/master/ports/telnetport.js

yaacov:
@LeonardoPilato ^^ ?

LeonardoPilato:
Thanks for your response and help, I'll close this pull request, I only posted here because I did not find another place for ask about Enron Modbus and the source is not my repository.
I'll work on do the necessary for Enron Modbus and I'll back with a correct pull request. Final question, my device for test use RTU over TCP, is the telnet port the right point for start to dev?

yaacov:
> my device for test use RTU over TCP, is the telnet port the right point for start to dev?

Yes, it transports RTU packets over TCP connection, you can also look at the RTU orer UDP port [1]

> I'll back with a correct pull request.

Nice 👍 

[1] https://github.com/yaacov/node-modbus-serial/blob/master/ports/c701port.js

Issue #159, opened 2017-11-01
serialport should be min v6.0.4


biancode:
Dear yaacov,

see [Issue](https://github.com/biancode/node-red-contrib-modbus/issues/61) early serialport versions has CPU load, which is fixed if the test is running with v6.0.4 of serialport package

Happy coding!
Klaus

yaacov:
can you make a pull request ?

biancode:
Yes sure, but what about v7.0.0 is it ready to use?

yaacov:
> Yes sure, but what about v7.0.0 is it ready to use?

Yes :-)

The changes where too small, so I waited with publishing a new release, It include the needed change from boudrate to baudRate, and the open issues are not blockers.

yaacov:
closed by #160

yaacov:
@biancode v7.0.1 is on npm 🍰

PR #160, opened 2017-11-01
minimum version 6.0.4 because of improved CPU load

biancode:
fix for issue #159


yaacov:
I will make a new release and publish it

PR #169, opened 2018-01-08
avoid double callback when read timeout is hit

dom-white:
In the case where an Ethernet cable is pulled out before a read there will be a callback of the read function with an Error (whenthe timeout code fires). 

However if the cable is re-plugged, due to the nature of tcp (reliable transport/re-transmissions), data may now arrive and an additional callback for the same read can be generated.

It is not good practice to call a callback multiple times, and if the read was issued within an async module task (e.g. forEachOfSeries), async will throw an exception due to the double callback.

This patch adds detection of whether the timeout fired for the transaction and prevents any data received for this transaction 'after' this to be dropped rather than sent back via a 2nd call to the callback.

Ideally this data would be usable, but I don't think there is a way of getting this back to the callee without triggering a double callback. Plus this is an edge condition where the link has dropped, so loosing a little bit of data would be acceptable.

### DIFF
--- a/index.js
+++ b/index.js
@@ -149,7 +149,8 @@ function _writeBufferToPort(buffer, transactionId) {
 
     this._port.write(buffer);
     if (transaction) {
-        transaction._timeoutHandle = _startTimeout(this._timeout, transaction.next);
+        transaction._timeoutFired = false;
+        transaction._timeoutHandle = _startTimeout(this._timeout, transaction);
     }
 }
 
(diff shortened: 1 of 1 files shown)


yaacov:
👍 Thanks

dom-white:
Thanks for merging this in. It all looks good to me now. Is it possible to create a new npm release (again :-) )

yaacov:
7.0.3 is out

dom-white:
brilliant. thanls

PR #170, opened 2018-01-11
Async server

uwburn:
Added support for callbacks (last function argument) and promises (if returned value has a then property).


yaacov:
Thanks, wow this is big, I will look into it this weekend.

p.s.
feel free to nag me, if you see I didn't ... :-)

yaacov:
Thanks @uwburn

Issue #163, opened 2017-11-08
Make serverTcp handler async?


meisterlampe:
I think about getting the register values from an async api and I don't know how to block the synchronous function until the promise is resolved.

After a first look into the code (servertcp.js), it seems possible to make `_parseModbusBuffer` async and await all the handlers.
What's your opinion? Is there a reason for the tcp server functions to be sync?

yaacov:
hi, thanks for the issue, I will look at it this weekend ... ( hopefully :-) )

meisterlampe:
In my case I just needed an async call to `getHoldingRegister`, so I can do some async work in the handler.
You can find my changes in https://github.com/meisterlampe/node-modbus-serial/blob/master/servers/servertcp.js

I know, it is a hack breaking the API and far away from beeing a PR as all other methods should still need to be converted to async. Anyway it's a working test for me right now :)
Would you prefer another (maybe non-api-breaking) approach?

yaacov:
wow, thanks, I hope to have time to look at it ...

uwburn:
Hello,
i'm working on a callback based non API breaking version of this, as i needed to simulate some delay of the network for testing purposes.

If it works out ok, i will write up tests for it and submit PR. ;)

yaacov:
great 👍

yaacov:
@meisterlampe please check  #170

meisterlampe:
Hm, I tried the current master branch and it was not working for me.
Maybe I'm using this wrong.

My test:
1. Using Smply Modbus TCP Client 8.0.7 to read a 32bit Float (working with my async mod mentioned in this issue)
2. Modbus-Server sends `4369 3A32`
3. Receiving only the first UInt16. Complete result is e.g.: `4369 0000`
4. Reading another 32bit Float
5. From now on the **previous** results are returned, not the current. E.g. `4369 3A32`

What I changed in my code:
Added callback to my vector, that is passed to `new ServerTCP`-call
`getHoldingRegister: (addr, unitId) => self.getHoldingRegister(addr, unitId)`
result:
`getHoldingRegister: (addr, unitId, callback) => self.getHoldingRegister(addr, unitId, callback),`

getHoldingRegister has this implementation:
```
private async getHoldingRegister(addr, unitId, callback) {
    try {
      Logger.debug(`getHoldingRegister: addr:'${addr}', unitID:'${unitId}'`);

      let result = await this.holdingRegisterManager.getHoldingRegister(addr);
[...]

PR #172, opened 2018-01-16
Fixed merge errors.

uwburn:


yaacov:
see: https://github.com/yaacov/node-modbus-serial/issues/163

will merge when green

PR #173, opened 2018-01-17
Only call callback when all parts received

meisterlampe:
As written in #163: This makes my code work well with a client reading a 32bit float (=2x UInt16).


yaacov:
Thanks @meisterlampe 👍

PR #176, opened 2018-01-19
Server error handling

uwburn:


yaacov:
Nice !

it's too big for me to read ... not this weekend anyway.

when tests will be green, I will merge it, and we will ask @meisterlampe to check if it doesn't break anything on his side, unless you want to do some more tests first ...

PR #185, opened 2018-02-09
Fixed TypeError when calling .open the second time

robertsLando:
Moved the isOpen defineProperty outside the .open method.


yaacov:
Thanks !

PR #189, opened 2018-02-27
Erros may be longer the 5 bytes

yaacov:
Some devices may send an error message that is longer then 5 bytes.


yaacov:
cc @biancode

PR #192, opened 2018-03-12
adding timeout to tcp connect, if it's set before connect is called

eeerin:


yaacov:
Thanks :+1:

yaacov:
Do you need this in a new npm version ?

eeerin:
yes please! 👍

yaacov:
:cake:  :sparkles:  :tada: 

https://github.com/yaacov/node-modbus-serial/releases/tag/v7.2.0
https://www.npmjs.com/package/modbus-serial

yaacov:
:cake: :sparkles: :tada: x 2 :-)

https://github.com/yaacov/node-modbus-serial/releases/tag/v7.2.1
https://www.npmjs.com/package/modbus-serial

uwburn:
Sorry to bother, but isn't this altering the previous behaviour?

The changes seems to call `socket.setTimeout(timeout[, callback])`, so, if there's not activity for a given timeout the socket will emit `timeout` event and the connection will be closed. The previous behaviour, if i understood correctly, was to consider the timout on answering a request.

I'm querying a Modbus device every minute or so, but i have a timeout as request not answered in one second at max indicate that probably something has gone wrong and it's better to recconect or to restart the application. With the latest version, after the initial query is successfull, the socket closes and i start getting errors/detecting the port has closed.

eeerin:
@uwburn this timeout will only be set if you call the function `client.setTimeout(1000);` on the client before making the connection. Try doing it after the connection is established. And that will give you the previous behavior. 
I dont think this should fire if you are making requests since the socket would have activity. I have been using this function and making a large number of requests serially and it works fine for me. Are you running the requests in parallel? Just curious when I was doing that, I would get one or two response and after that the socket would hang. This was my solution towards getting a responable response when 1. the socket connection fails to open in a given amount of time and 2. when the socket is just sitting there doing nothing for too long.

uwburn:
Ok, i've understood the problem: i have a routine that's checking if the connection with the client becomes unresponsive or if the port has been closed: in that case i call connect again on the client, so that the connection is restabilished. In that case it's just like it was called before connecting. I see if i can work around it, thanks for the clarification. ;)

eeerin:
No worries! Always glad to help! 😄

PR #195, opened 2018-03-12
add timeout to tcpbufferd and telnet ports

yaacov:
in #192 we add timeout also to tcp port, here we add timeout to tcpbuffered and to telnet ports


yaacov:
cc @eeerin

PR #197, opened 2018-03-13
adding destroy functions to each net.Socket ports

eeerin:
After testing the timeouts further, I noticed that if you got the timeout signal the socket will continue trying to connect, because calling `socket.end` only half closes the socket trying to send a `fin` packet. However calling destroy kills the socket completely. I looked through the serial-ports docs and did not find an a destory function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -375,6 +375,20 @@ ModbusRTU.prototype.close = function(callback) {
 };
 
 /**
+ * Destory the serial port
+ *
+ * @param {Function} callback the function to call next on close success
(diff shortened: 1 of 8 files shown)


yaacov:
Nice, thanks !

yaacov:
7.2.2 is out :tada:

eeerin:
@yaacov if you could republish that would be awesome! Thank you!

eeerin:
You're too quick for me! 😄

PR #209, opened 2018-06-04
Timeout on transactions

wouldgo:
Hi,

we're using the library to connect to a [Moxa NPort 5130](https://www.moxa.com/product/nport_5130.htm).
During a maintenance by pure chance we disconnect the serial port and the software did not reply with any kind of error or notification on the next reading round (we read on a given interval the values inside some registers).

I introduced the timeout also on transactions using the same value passed on the connect. In this way our software get notified by the timeout error and the reading round is ignored.

We consider this valuable to be in the library. I honestly did not understand if there were any other ways to achieve it.

Thanks for the commitment on the library. 
Hope to brought value,

have a great day,
Dario

### DIFF
--- a/apis/connection.js
+++ b/apis/connection.js
@@ -106,6 +106,8 @@ var addConnctionAPI = function(Modbus) {
         var TcpPort = require("../ports/tcpport");
         if (this._timeout) {
             options.timeout = this._timeout;
+        } else if (options.timeout) {
+            this._timeout = options.timeout;
         }
         this._port = new TcpPort(ip, options);
 
@@ -135,6 +137,8 @@ var addConnctionAPI = function(Modbus) {
         var TcpRTUBufferedPort = require("../ports/tcprtubufferedport");
         if (this._timeout) {
             options.timeout = this._timeout;
+        } else if (options.timeout) {
+            this._timeout = options.timeout;
         }
         this._port = new TcpRTUBufferedPort(ip, options);
 
@@ -165,6 +169,8 @@ var addConnctionAPI = function(Modbus) {
         var TelnetPort = require("../ports/telnetport");
         if (this._timeout) {
             options.timeout = this._timeout;
+        } else if (options.timeout) {
+            this._timeout = options.timeout;
         }
         this._port = new TelnetPort(ip, options);
 


dushyantbangal:
There IS a timeout that is put on all the requests.
I've only tried it with TCP, so I'm not completely sure about RTU.

After the connection is done, you just need to
     client.setTimeout(NUMBER_OF_MILLISECONDS);
and then you will get a timeout error when the server doesn't respond in the specified time.

https://github.com/yaacov/node-modbus-serial/blob/master/examples/simple.js#L28

yaacov:
Hi, @wouldgo Thanks for the code, happy to get pull requests :heart:  

As @dushyantbangal commented we have a timeout -
A. Are you sure the current implementation of timeout is not working ?
B. Do you know why the current implementation does not work ?
https://github.com/yaacov/node-modbus-serial/blob/master/ports/tcpport.js#L46

Is it a bug ? should we fix that ?

p.s.
The `_timeout` that you are setting is using an internal mechanism for timouts:
https://github.com/yaacov/node-modbus-serial/blob/master/index.js#L165
That does not care about the client implementation of "timeout". so it will work when the
clients timeout fails ...

It's a good solution if the clients timeout does not fire, but I want to be sure about that ...
https://github.com/yaacov/node-modbus-serial/blob/master/ports/tcpport.js#L46

wouldgo:
Hi,

did not get the `setTimeout` method (I search on the codebase briefly before doing what I did). I'll try and I'll let you know.

wouldgo:
@yaacov My scenario is this one: 
```const ModbusRTU = require('modbus-serial');

return client.connectTelnet(<address>, {
    'port': <port>,
    'timeout': 3000
  })
  .then(() => {

    client.setID(1);
    return client.readHoldingRegisters(3029, 8);
  }, err => console.info('connect err', err))
  .then(data => console.info(data), err => console.info('reading err', err))
  .then(() => process.exit(), err => console.info('data err', err))
  .catch(err => {

    console.error(err);
  });
```

If I put `client.setTimeout(3000);`, as suggested by @dushyantbangal, before the `client.setID(1)` all it's working as expected (and so I think that my PR is useless).

yaacov:
@wouldgo  and @dushyantbangal :heart: thnaks for raising this issue.

> If I put client.setTimeout(3000);, as suggested by @dushyantbangal, before the client.setID(1) all it's working as expected (and so I think that my PR is useless).

Thanks even if we can not use this fix as is :+1: , I did not know about this problem, and we do need to fix this, and understand why the clients timeout does not fire [1] -
a. to document this.
b. to see why the timeout from the tcp client does not fire ... maybe we do not need it (if it does not fire on timeouts)

[1] https://github.com/yaacov/node-modbus-serial/blob/master/ports/tcpport.js#L98

yaacov:
@wouldgo 

a. is it ok to close this one ?
b. do you want to make it a documentation (add timeout example + a line in the readme) fix, so others will know how to set the timeout ?

yaacov:
@wouldgo closing. we will need to document this timout issue, so others will not have the same problem.

wouldgo:
@yaacov Sorry for the late reply, Your're right about closing this PR. Just to recap: my issue with timeout was related to the RTU scenario.

How do you prefer have a working example and a documentation? In this way I can focus on this as soon as I can.

Dario.

yaacov:
> How do you prefer have a working example and a documentation? In this way I can focus on this as soon as I can.

I will be very happy if we have a working example, and after it a section in the `readme` with reference to this example.

PR #211, opened 2018-06-04
Fixes #130

### LINKED ISSUE #130: ServerTCP event
Hi,

```
var vector = {
    getInputRegister: function(addr, unitID) { return addr; },
    getHoldingRegister: function(addr, unitID) { return addr + 8000; },
    getCoil: function(addr, unitID) { return (addr % 2) === 0; },
    setRegister: function(addr, value, unitID) { console.log("set register", addr, value); return; },
    setCoil: function(addr, value, unitID) { console.log("set coil", addr, value); return; }
};
```
`var serverTCP = new ModbusRTU.ServerTCP(vector, { host: "0.0.0.0", port: 8502, debug: true });`

For example, the modbus client TCP sends a request to read 10 holdingRegs (addr from 0-9)
What I tested is the vector will be called 10 times continuously, each time for each addr (0, 1, ..., 9) and return the corresponding value.

**My issues are:**
- How to know when will the client modbus send a request? meaning how to get that event to control other functions? Because I want other functions to be implemented right after the serverTCP got the request from Client. And those functions should be implemented before going to vector and respond to Client.
- How to get the entire package from Client Modbus? Because sometime I want to know how many registers the client requests, and the range of address (from 0-9 in this example, but sometime it will be from addr 5 to addr 14...), and maybe sometime I want to know the unitID in the client's package, currently I go to vector variable to get the unitID but I think that's not a good way.
So is there any ways to get all parameters in client's message: the range of address, unitID...

**P/S:** I'm a beginner in Node js, so maybe my questions are so basic, please kindly help me if you don't mind :D

Thanks for your time!

dushyantbangal:
Added two new functions to vector:
1. `getMultipleInputRegisters`
2. `getMultipleHoldingRegisters`

If defined, these functions will be called once for each request. These functions should return an array of values. If the length of the returned array does not match the length of the requested number of register values, and Error is thrown and Exception 4 is sent.

If not defined, the old technique of calling `getInputRegister` and `getHoldingRegister` multiple times is used. This has been done in case someone still wants to use the old technique or has the old code but has updated package.

### DIFF
--- a/examples/server.js
+++ b/examples/server.js
@@ -6,17 +6,17 @@ var ModbusRTU = require("../index");
 var vector = {
     getInputRegister: function(addr) { return addr; },
     getHoldingRegister: function(addr) { return addr + 8000; },
-    getMultipleInputRegisters: function (startAddr, length){
-        var values=[];
-        for(var i=0;i<length;i++){
-            values[i]=startAddr+i;
+    getMultipleInputRegisters: function (startAddr, length) {
+        var values = [];
+        for (var i = 0; i < length; i++) {
+            values[i] = startAddr + i;
         }
         return values;
     },
-    getMultipleHoldingRegisters: function (startAddr, length){
-        var values=[];
-        for(var i=0;i<length;i++){
-            values[i]=startAddr+i+8000;
+    getMultipleHoldingRegisters: function (startAddr, length) {
+        var values = [];
+        for (var i = 0; i < length; i++) {
+            values[i] = startAddr + i + 8000;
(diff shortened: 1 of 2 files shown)


yaacov:
Nice ! , can you fix the linting errors ?

dushyantbangal:
Can you provide a rule file so I can check before commit?

dushyantbangal:
Done

yaacov:
@dushyantbangal Thanks :+1:

PR #215, opened 2018-06-12
add explicit overrides for promise vs callback connections

reidev275:
The code within the `open` function https://github.com/yaacov/node-modbus-serial/blob/master/apis/connection.js#L33-L51 checks to see if a callback is provided and if not returns a promise.  This PR updates the index.d.ts to reflect that.


yaacov:
@reidev275 Thanks, will merge when green !

please comment if you want to change something before I merge.

PR #239, opened 2018-10-17
typescript import fix

sh84:
Now 
`import ModbusRTU from 'modbus-serial';`
raise error
TypeError: modbus_serial_1.default is not a constructor

This small fix add default to exports.

### DIFF
--- a/index.js
+++ b/index.js
@@ -734,6 +734,7 @@ require("./apis/promise")(ModbusRTU);
 
 // exports
 module.exports = ModbusRTU;
+module.exports.default = ModbusRTU;
 module.exports.TestPort = require("./ports/testport");
 try {
     module.exports.RTUBufferedPort = require("./ports/rtubufferedport");


yaacov:
@mainakae  can you take a look ?

Why do we need change the javascript code , if the problem is in the typescript ?

mainakae:
Hi!

I think the solution proposed by @sh84 is unnecessary, as the typescript compiler generates this code at the beginning of the file:

```javascript
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
```
which encapsulates the result of require in the "default" property, thus rendering the patch unnecessary.
I maybe @sh84 is having some problem with his TS compiler? maybe configuration options?, which TS compiler version are you using?

I'm using tsc version 3.1.1, npm 5.6.0 and node 8.11.3

My current TS config is:
```javascript
{
  "compilerOptions": {
    "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
    "strict": true,                           /* Enable all strict type-checking options. */
    "esModuleInterop": true                   /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
  },
  "include": [
    "./src/**/*"
  ],
  "exclude": [
    "./src/**/*.spec.ts"
  ]
}

sh84:
You are right, my problem is esModuleInterop=false.

yaacov:
@sh84 Thanks for the PR :+1: 

Next person that will fail because of `esModuleInterop` settings, will have this PR as a reference.

yaacov:
@mainakae Thanks for the quick answer :100:

mainakae:
No prob :ok_hand: :smiley:

Issue #241, opened 2018-11-06
Uncaught TypeError: "value" argument is out of bounds


w4tsn:
I'm implementing a TCP server using this library. When I'm trying to read any registers from e.g. 0 to 200, requesting 200 registers, a request buffer overflow is thrown and uncaught. This causes my application to crash, since the error is thrown outside of my control flow within this library. My vector of request methods is thus never called.

I identified a solution. Is it desired that I file a pull-request for this, or is there something I do not see here? May this even be desired? Is there a configuration to tackle this or is it even a modbus specific thing?

The error is thrown in each `servertcp_handler.js` method, where `writeUInt8()` is called with a requested buffer length greater than what `writeUInt8` buffers support.

yaacov:
> I identified a solution. Is it desired that I file a pull-request for this,

a. Yes :+1: 
b. This is tricky since you should not break the Serial RTU struct that is of UInt8 length.

PR #242, opened 2018-11-08
fixes issue #241

w4tsn:
This PR fixes an issue where the modbus-serial library would throw a unhandled exception if a client makes a request for a large amount of registers at once. The basic idea is, that this exceeds the write calls to the UInt8 buffers. To tackle this, these calls are encapsulated in try/catch blocks and a slave error of 0x04 is returned to the client.

I decided to not throw this further to the `ServerTCP._parseModbusBuffer()` function for output through the debug function, since these errors may be best handled directly in the handler functions. It may be desired to do this in the future though.

Tests included.

### DIFF
--- a/servers/servertcp_handler.js
+++ b/servers/servertcp_handler.js
@@ -80,7 +80,13 @@ function _handleReadCoilsOrInputDiscretes(requestBuffer, vector, unitID, callbac
     // build answer
     var dataBytes = parseInt((length - 1) / 8 + 1);
     var responseBuffer = Buffer.alloc(3 + dataBytes + 2);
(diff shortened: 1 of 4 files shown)


yaacov:
Thanks :+1: 

p.s
Do you need a new version in npm ?

yaacov:
+1 for the tests :100:

w4tsn:
No problem and yes, a new version would be great!

yaacov:
> a new version would be great!

:tada: :cake: 

https://www.npmjs.com/package/modbus-serial/v/7.4.1

Issue #243, opened 2018-11-10
Module fails to build via TypeScript


ericbarch:
Greetings!

It appears that the latest release (7.4.1) no longer compiles properly under TypeScript. My projects that use modbus-serial now fail with:

`node_modules/modbus-serial/ServerTCP.d.ts:1:8 - error TS1192: Module '"events"' has no default export.`

Note that I am actually using ModbusRTU with the default export: `import ModbusRTU from 'modbus-serial';`.

yaacov:
@ericbarch can you look at:

https://github.com/yaacov/node-modbus-serial/pull/239#issuecomment-430650820

Is it related to your problem ?

ericbarch:
Just opened PR #244. This should fix it 😃

ericbarch:
I did take a look at #239, but I am using the default export (which works fine). The issue reported here is related to incorrect syntax of an ES6 import via TypeScript.

yaacov:
closed via #244

PR #244, opened 2018-11-10
Use proper ES6 import syntax for TypeScript types

ericbarch:
The built-in events NodeJS module does not have a default export. This prevents modbus-serial from compiling (as of release 7.4.1).

By using the proper ES6 import syntax, the module is imported without error.


yaacov:
@ericbarch Thanks !

yaacov:
Fixes #243

yaacov:
new version in npm :tada:

Issue #240, opened 2018-10-18
read/write fail after reconnecting RTU


dushyantbangal:
I am using a RS485-USB converter to connect to meter.
I am on Windows 10, node 8, latest package version from npm.

I have a simple write command followed by a read command, both wrapped in a `setInterval`
Before the commands, I check if the port if open or not. If its not, then I start reconnection. The connection is retried every 1500 ms.

If I unplug the converter and plug it back once it starts reconnection, the connection is successful. However, most of the times, the requests keep failing, until I kill the script and restart it again.

```
var ModbusRTU = require("modbus-serial");
var client = new ModbusRTU();
var interval;

connect();

function write() {

    if(!client.isOpen)
    {
        clearInterval(interval);
        connect();
        return;
    }

    console.log("sending write");
    client.setID(1);

    client.writeRegister(5, 2).then(function(resp){
        console.log(resp);
        read();
    }).catch(function(err){
        console.log(err);
    })
        
}

function read() {
    console.log("sending read");

    client.readHoldingRegisters(5, 1)
        .then(console.log)
        .catch(console.log);
}


function connect(){
    client.connectRTU("COM5", { baudRate: 9600, parity: 'none', dataBits:8, stopBits:1 }, function(err){
        if(err){
            console.log(err);
            return setTimeout(connect,1500);
        }else{
            console.log("connected");
            interval = setInterval(write, 5000);
        }
    });
}
```

Output:

```
connected
sending write
{ address: 5, value: 2 }
sending read
{ data: [ 2 ], buffer: <Buffer 00 02> }
sending write
{ address: 5, value: 2 }
sending read
{ data: [ 2 ], buffer: <Buffer 00 02> }        //UNPLUG
Error: Opening COM5: File not found
Error: Opening COM5: File not found
Error: Opening COM5: File not found            //PLUG
connected
sending write
{ address: 5, value: 2 }
sending read
sending write
{ address: 5, value: 2 }
sending read
sending write
{ address: 5, value: 2 }
sending read
```

yaacov:
@dushyantbangal Thanks :+1: , I hope to have time to look at it this weekend, but I have very little time ...
If you can investigate more / find a solution it will be great.

gabrielpereirapinheiro:
I'm having the same problem ! Someone find the soluction and can asnwer here  ?!?

gabrielpereirapinheiro:
I'm having the same problem !
Someone find the solution and can answer here ? Thanks

[...]

PR #254, opened 2019-01-24
Add framework examples #240

sigma-design:
As requested in #240 , I have added framework examples of polling Modbus holding registers.

Please don't pay attention to multiple commits of the same files - I forgot to give a proper title/message, so I have committed (same) files again to be able to change the message (maybe there is an easier way, but I could not see it immediately).

### DIFF
--- a/examples/polling_RTU.js
+++ b/examples/polling_RTU.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-console, spaced-comment */
 
 //==============================================================
-// This is an example of polling (reading) Holding Registers
+// This is example of polling (reading) Holding Registers
 // on a regular scan interval with timeouts enabled.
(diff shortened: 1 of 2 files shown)


yaacov:
:+1:  Thanks !  :cake:

yaacov:
> Please don't pay attention to multiple commits of the same files

I'm not :-)

p.s.
The git command you are missing is:
``` bash
git rebase -i HEAD~[n] # n is the number of commits to sqash
```

this will open an editor where you will choose what to do with the last `n` commits, see:
http://www.andrewconnell.com/blog/squash-multiple-git-commits-into-one

sigma-design:
I have used browser for forks/commits...
Next time I will check if browser based commits can be squashed as well.

Issue #247, opened 2018-12-28
FC16 handling and Remote Address


tizl1234:
Hi,
I have two questions:
 1) How can I handle FC16 on modbus TCP server? If I use such code [example](https://jsfiddle.net/tizl1234/5ofn1ytx/3/) I receive values one by one. It's not apliable in my case. It would be better if i could receive values in array format. Maybe there is a way to do that in this module?
2) How can I receive remote address of slave device in my serverTCP vector?

Thank you!

yaacov:
> How can I handle FC16 on modbus TCP server? If I use such code example I receive values one by one. It's not apliable in my case. It would be better if i could receive values in array format. Maybe there is a way to do that in this module?

No, the function that implement multiple write is here, it just loop over the values and call the single-set method:
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp_handler.js#L602

Multiple reads **are** implemented, you can look at the code that handle multiple reads here:
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp_handler.js#L221
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp_handler.js#L255

You can make a pull request that does the same thing for the write method.

> How can I receive remote address of slave device in my serverTCP vector?

No,

The function that calls the server vector methods is:
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp.js#L125

You can add the remote adress when calling it:
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp.js#L271

using:
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp.js#L218

and then propagate it to the vector calls.

Will be happy to here if that helped, have fun.

PR #260, opened 2019-03-05
FC15 and FC16 handlers pass values as single array

tizl1234:
#247 Added two new function to vector:

1. `setRegisterArray`
2. `setCoilArray`

If defined, these functions pass values as a single array, instead of passing them on by one.
If you define `setRegister` and `setRegisterArray` simultaneously the old technique will work and the new one will be ignored (same for the coils).


yaacov:
Hi, Thanks !
Waiting for tests to pass :-)

yaacov:
Thanks !

yaacov:
That's great, thanks, have fun :beers:  :cake:

p.s.
If you have some time, please add tests, so it will not get broken accidentally in the future.

PR #267, opened 2019-05-03
Use callbacks in close() and destroy() even if nothing needs to be done

adam-nielsen:
In my code I call `close()` after getting an error to attempt a reconnect, but sometimes the port is already closed when this happens.  In this situation the callback never gets called, so my code freezes waiting forever for the port to close.

By calling the callback even when nothing needs to be done, the code will no longer get stuck.

### DIFF
--- a/index.js
+++ b/index.js
@@ -380,6 +380,9 @@ ModbusRTU.prototype.close = function(callback) {
     if (this._port) {
         this._port.removeAllListeners("data");
         this._port.close(callback);
+    } else {
+        // nothing needed to be done
+        callback();
     }
 };
 
@@ -394,6 +397,9 @@ ModbusRTU.prototype.destroy = function(callback) {
     if (this._port && this._port.destroy) {
         this._port.removeAllListeners("data");
         this._port.destroy(callback);
+    } else {
+        // nothing needed to be done
+        callback();
     }
 };
 


yaacov:
Hi, thanks :+1:

yaacov:
Do you need a new version in npm ?

adam-nielsen:
Would be nice eventually but no rush at the moment - have just copied the file across into `node_modules` manually.  Thanks for merging!

yaacov:
Thanks, have fun weekend :cake:

sacha67:
Hello, do you know when you will release a new version with this fix ?

 I have the same problem.
`
				const closePromise: Promise<void> = new Promise((resolve): void => {
					connection.modbusRTU.close(() => resolve());
				});
`
The promise encapsulation doesn't always gets resolved and freezes my app. :'(

yaacov:
> Hello, do you know when you will release a new version with this fix ?

Ping me again here before the weekend :peace_symbol:

sacha67:
Ping before the weekend 😄

yaacov:
@sacha67 
https://www.npmjs.com/package/modbus-serial :tada: :cake:

sacha67:
Thank you ! 🍰

adam-nielsen:
@sacha67 Just FYI this patch only seems to fix the problem some of the time.  There is still somewhere that does not correctly trigger the callback so even with my patch occasionally the promise won't resolve.

I still haven't been able to track down where this happens, but it looks like it may be an error in one of the other libraries and not `node-modbus-serial`.

sacha67:
@adam-nielsen Thank you for the warning.

PR #270, opened 2019-05-20
ServerTCP initialized event added

onur2677:


servers/servertcp.js
```diff
@@ -207,7 +207,12 @@ var ServerTCP = function(vector, options) {
 
     // create a tcp server
     modbus._server = net.createServer();
-    modbus._server.listen(options.port || MODBUS_PORT, options.host || HOST);
+    modbus._server.listen({
+        port : options.port || MODBUS_PORT,
+        host: options.host || HOST
+    }, function() {
+        modbus.emit('initialized')
```
yaacov: in :
https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp.js#L218

We use the notation `server.on("connection", [ init_function ] `
here you use the notation of  `server.listen(..., [ init_function ] `

I think it is nicer to use only one notation :-) ... my vote is to use the current notation and move the `emit` to line 220, inside the current init function.

onur2677: ```server.listen()```  and ```server.on("connection")``` for different purposes.
I want to know when the server was started.

server.listen :  [When the server starts listening, the 'listening' event will be emitted. The last parameter callback will be added as a listener for the 'listening' event.](https://nodejs.org/api/net.html#net_server_listen)

server.on("connection"):  [Emitted when a new connection is made](https://nodejs.org/api/net.html#net_event_connection)

yaacov: nice :+1:

yaacov:
can you fix the linting ?
https://travis-ci.org/yaacov/node-modbus-serial/jobs/534709380

onur2677:
oh my bad :)

yaacov:
Thanks :exclamation:

PR #277, opened 2019-07-25
Added support to concatenate multiple data events into one response

coder-mike:
I was running into a problem with low baud rates, where the serial port was splitting data at unexpected boundaries, so the `"data"` serial port event was firing multiple times with different pieces of what should be a single response.

Example as follows, with a bulk request for 61 registers over a 19200 baud connection.

```
TX Buffer(8) [13, 3, 32, 0, 0, 61, 143, 23]
RX  Buffer(23) [13, 3, 122, 0, 200, 27, 0, 0, 1, 0, 3, 0, 0, 0, 0, 4, 172, 0, 6, 0, 0, 7, 208]
RX  Buffer(31) [7, 218, 0, 249, 0, 248, 0, 0, 0, 0, 0, 0, 0, 0, 7, 208, 7, 218, 0, 249, 0, 248, 0, 0, 0, 0, 0, 0, 0, 0, 0]
RX  Buffer(31) [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 255, 255, 0, 0, 0, 0, 255, 255, 255, 255, 255, 255]
RX  Buffer(31) [255, 255, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 244, 37, 28, 0, 50, 32, 8, 38, 72, 0, 10, 0, 0, 12, 56, 32]
RX  Buffer(11) [10, 0, 0, 14, 251, 1, 94, 0, 18, 53, 254]
```

This change allows multiple data events to be concatenated together before processing. If the required data never arrives, a timeout is issued as normal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ var ModbusRTU = require("modbus-serial");
 var client = new ModbusRTU();
 
 // open connection to a serial port
-client.connectRTU("/dev/ttyUSB0", { baudRate: 9600 }, write);
+client.connectRTUBuffered("/dev/ttyUSB0", { baudRate: 9600 }, write);
 
 function write() {
     client.setID(1);
--- a/examples/buffer.js
+++ b/examples/buffer.js
@@ -6,7 +6,7 @@ var ModbusRTU = require("../index");
 var client = new ModbusRTU();
 
 // open connection to a serial port
-//client.connectRTU("/dev/ttyUSB0", {baudRate: 9600})
+//client.connectRTUBuffered("/dev/ttyUSB0", {baudRate: 9600})
 client.connectTCP("127.0.0.1", { port: 8502 })
     .then(setClient)
     .then(function() {
--- a/examples/debug.js
+++ b/examples/debug.js
@@ -15,7 +15,7 @@ var ModbusRTU = require("../index");
 var client = new ModbusRTU();
 
 // open connection to a serial port
-//client.connectRTU("/dev/ttyUSB0", {baudRate: 9600})
+//client.connectRTUBuffered("/dev/ttyUSB0", {baudRate: 9600})
 client.connectTCP("127.0.0.1", { port: 8502 })
     .then(setClient)
     .then(function() {
--- a/examples/simple.js
+++ b/examples/simple.js
@@ -8,7 +8,7 @@ var client = new ModbusRTU();
 var networkErrors = ["ESOCKETTIMEDOUT", "ETIMEDOUT", "ECONNRESET", "ECONNREFUSED", "EHOSTUNREACH"];
 
 // open connection to a serial port
-client.connectRTU("/dev/ttyUSB0", { hupcl: false, dsrdtr: false })
+client.connectRTUBuffered("/dev/ttyUSB0", { hupcl: false, dsrdtr: false })
 //client.connectTCP("modbus.local", { port: 502 })
     .then(setClient)
     .then(function() {
--- a/examples/write.js
+++ b/examples/write.js
@@ -6,7 +6,7 @@ var ModbusRTU = require("../index");
 var client = new ModbusRTU();
 
 // open connection to a serial port
-//client.connectRTU("/dev/ttyUSB0", {baudRate: 9600})
+//client.connectRTUBuffered("/dev/ttyUSB0", {baudRate: 9600})
 client.connectTCP("127.0.0.1", { port: 8502 })
     .then(setClient)
     .then(function() {
--- a/examples/write_complete.js
+++ b/examples/write_complete.js
@@ -6,7 +6,7 @@ var ModbusRTU = require("../index");
 var client = new ModbusRTU();
 
 // open connection to a serial port
-// client.connectRTU("/dev/ttyUSB0", {baudRate: 9600})
+// client.connectRTUBuffered("/dev/ttyUSB0", {baudRate: 9600})
 client.connectTCP("127.0.0.1", { port: 8502 })
     .then(setClient)
     .then(function() {


yaacov:
did you try the buffered port ?

https://github.com/yaacov/node-modbus-serial/blob/master/apis/connection.js#L209
https://github.com/yaacov/node-modbus-serial/blob/master/ports/rtubufferedport.js

coder-mike:
Shouldn't the "normal" port provide whatever implementation is required to support modbus? The "data" event on the standard SerialPort is not guaranteed to deterministically provide any specific length of data or break at specific points, so it can't reliably be used without any form of buffering.

yaacov:
> The "data" event on the standard SerialPort is not guaranteed to deterministically provide any specific length of data or break at specific points, so it can't reliably be used without any form of buffering.

Actually it should, https://en.wikipedia.org/wiki/Modbus#Frame_format.

But since we know the real world is not that nice ( as you noticed ), we have the buffered port, witch is very similar to the code you suggest :-)

yaacov:
@coder-mike Thanks for the pull request :heart: , I do not want to add logic in the main modbus packet parser file, this logic can be in one of the ports.
I think that the buffer port has similar logic, that may fix the issue you raised :+1: 

If the buffered port solve you problem, we can:

a - add some more documentation in:
https://github.com/yaacov/node-modbus-serial/blob/master/README.md
to help the next people having this problem.

b - add an example code in:
https://github.com/yaacov/node-modbus-serial/tree/master/examples
that illustrate how to use the buffered port.

c - add something about it in the wiki.

if the bufferd-port does not help, we should fix it :smile:

coder-mike:
>> The "data" event on the standard SerialPort is not guaranteed to deterministically provide any specific length of data or break at specific points, so it can't reliably be used without any form of buffering.
>
> Actually it should, https://en.wikipedia.org/wiki/Modbus#Frame_format.

I have some experience writing applications using native serial ports on both Windows and Linux. It's possible to configure the native serial port to either break received frames when there is a certain time delay between bytes, or when a certain read size is reached (i.e. the receive buffer is filled up). I don't know whether the npm `serialport` wrapper library exposes these or not, but either way, it does not look like the `modbus-serial` is leveraging any kind of control over these parameters (correct me if I'm wrong) in order to correctly implement the modbus protocol.

What I believe this means is that the `"data"` events coming from the `SerialPort` are basically served as fast as the OS chooses to serve them, which is dependent on the OS process/thread scheduler. On my machine (Windows), I believe the scheduler works at a resolution of 15.4ms, which I think is why it's able to buffer about 31 bytes of data at 19200 baud before it considers the read operation to be completed at the OS level, and provides that to the SerialPort library and in turn the modbus-serial library.

This is not a bug in the SerialPort implementation: the SerialPort library provides no guarantee (that I'm aware of) that a long received stream of bytes is chunked in any particular way. So the fact that it happens to be chunked on a 15.4ms boundary on my machine is perfectly fine.

What I'm concerned about is that the modbus-serial library is written in a way that's dependent on the chunk size, even though the SerialPort does not provide any chunking guarantees (I'm using the term "chunk" as per the node stream API https://nodejs.org/api/stream.html). This means that it will work correctly by coincidence for short responses (small responses, or medium responses at high baud rate), but fail to correctly work with longer responses. Even with shorter responses, if your PC is under load or any number of other factors, it's perfectly valid for it to break the chunks even smaller and so you'll get random unexpected failures. Again, please correct me if the SerialPort _does_ document chunking guarantees, but is not upholding these guarantees (then I'll file a bug with the `serialport` NPM library). 

I do think that the buffered port fixes this problem -- I didn't know about it until you pointed it out, thank you (I haven't tested it, but the code looks like it achieves the chunk-merging that's required here). I do also think that documentation will help point people to the buffered port. 

But I strongly believe that if the standard "port" can't deliver consistent behavior with a standard modbus device, then it shouldn't be the default port used by the modbus-serial library, or it should be patched so that it does provide consistent behavior.

coder-mike:
I think this may be relevant:
```
// ReadFileEx and GetOverlappedResult retrieved only 1 byte. Read any additional data in the input
// buffer. Set the timeout to MAXDWORD in order to disable timeouts, so the read operation will
// return immediately no matter how much data is available.
```
(https://github.com/serialport/node-serialport/blob/master/packages/bindings/src/serialport_win.cpp#L480)

This says to me that when there is any data (i.e. at least one byte is received), whatever is in the buffer will be passed up to the `"data"` event handler. The key here is that there is no particular amount of data that is returned: as soon as the OS schedules the event corresponding to the first byte being received, all the data received so far will form part of the chunk (from the first byte, up to and including all the data received in the meantime).

Again, I think this behavior is correct and expected. The modbus-serial library should be written with this in mind (i.e. written with the fact that `"data"` does not necessarily hold the whole response).

yaacov:
Thanks the research ! I accept that we can not trust the OS to cut incoming data as expected :+1:

This PR has some problems:

Missing in this implementation:
1 - Exception messages have different size then the `nextLength`.
2 - Broken messages emanating from faulty device / line.
3 - Data after a message that is the beginning of a new message.

Design of code:
1 - This part of the code is common to TCP / RTU / ASCII / Telnet  and any other port implementation we may invent, some of this port have a "sliding window" / "partial messages" problem while other may not, I prefer the port code will handle that problem on the port level.

My problems with this code:
1 = The code is implemented in the wrong place (it should be implemented in the code handling data coming from the OS driver) and not the data coming from the library port. data coming from the port should be segmented correctly.
2 - It does not take into account negative flow scenarios, like legitimate exception and broken messages.
3 - We already have a port that handle broken messages coming from the OS driver (or range extenders / networking ... ).

I think the best solution is better documentation. Maybe promote the buffered RTU port to be the default one, WDYT ?

p.s.
@biancode , I think the buffered one is the default in node-red ?

coder-mike:
Thanks for the great feedback. Yeah, I see what you're saying. 

I can agree with a solution that does not document the non-buffered RTU connection as part of the public API, since it doesn't work in the general case, and instead document the buffered approach. I.e. the `connectRTUBuffered` call should be the one shown in the readme and examples, rather than `connectRTU`.

Is there any situation in which `connectRTU` would need to be used instead of `connectRTUBuffered`? I.e. is there any situation in which non-buffered communication can guarantee correct RTU modbus communication for any reasonable message length and provide functionality that the buffered communication doesn't? If not, then I might suggest that `connectRTU` simply be an alias for `connectBufferedRTU`. Otherwise, what about just changing the documentation so `connectRTU` is deprecated and the corresponding examples are changed accordingly so that people copy-pasting the examples have something that works right off the bat?

yaacov:
> the `connectRTUBuffered` call should be the one shown in the readme and examples, rather than `connectRTU`.

Sounds good to me :heart:

> Is there any situation in which `connectRTU` would need to be used instead of `connectRTUBuffered` ? 

Yes, it's the vanilla low level option, like any low level option it helps when you want to see what happen with the least noise from unexpected code paths.

coder-mike:
Sounds good! Would you like me to change the documentation (and issue a second pull request), or would you like to do it?

yaacov:
> Would you like me to change the documentation (and issue a second pull request),

Sure, that would be nice !

p.s. 
This PR failed on ci checks, on the next PRs,  make sure everything in green and passing :-)

coder-mike:
Oh I see, thanks for the tip. 

Since the next one will just be a documentation change, I assume it will just pass.

coder-mike:
I've made the changes. It was less work than I expected, but tell me if I've missed something. I was going to make it a separate pull request, but I see that when I force-pushed to my `master`, it's already showing up in this pull request.

yaacov:
Thanks !

PR #284, opened 2019-09-26
more refactor of as buffer

yaacov:
cc:// @skangmy


skangmy:
Good one 👍

BTW, thanks for this great package. I have benefited from this a lot! ✌

PR #292, opened 2019-10-16
Make recvBuffer part of the connection handler

proditis:
Not sure if this is the correct fix or if it breaks other parts of the modbus server but here it goes.

When more than one client connect to ServerTCP, the recvBuffer gets mangled and thus render the server incapable of processing any further requests. The problem persists even after the clients close their connection.

You can reproduce the problem by:

1. run the sample server
2. connect with telnet and press enter 
3. close telnet
4. try to read a holding register with a modbus client
5. the client gets stuck there

With the proposed change the problem gets fixed for my cases.


yaacov:
Thanks :heart:

PR #294, opened 2019-10-16
Fix var i beeing used before assignment

NicolasDuran:
Hi,

In your last patch, you introduced a bug that breaks getCoil and getDiscreteInput functions of the TCP server by using a variable before assigning it.
This merge request fixes it


yaacov:
Thanks !
will merge when green :heart:

PR #295, opened 2019-10-16
Fix callback bit offset FC1/FC2 beeing forced to 0

NicolasDuran:
FC1 and FC2 callback had a bit offset forced to 0.
If doing a FC1/FC2 request on multiple adresses, only the first result was set


yaacov:
whoo, nice ! I think I did that ...
will merge when green :-)

yaacov:
p.s.
if you have time, we need tests for this part of the code ... just if you have time :-)

yaacov:
@NicolasDuran need linting fix ^

NicolasDuran:
I will try to add tests if I have time in the next days :)

PR #296, opened 2019-10-21
Allow TCP connections to originate from specific IP addresses

adam-nielsen:
We've got two Modbus programs running on the same machine, and they are both querying a Siemens PLC via Modbus TCP.  Unfortunately despite the Siemens device being brand new, it is running an extremely limited IP stack and if it gets multiple connections from the same IP address it drops all traffic.

The only solution is to ensure there is only one connection per IP address, which means we have to add two IP addresses to the machine, and then tell this library which IP address to use for outgoing packets.  This PR implements that feature.

### DIFF
--- a/ports/tcpport.js
+++ b/ports/tcpport.js
@@ -19,18 +19,25 @@ var CRC_LENGTH = 2;
  *
  * @param ip
  * @param options
+ *   options.port: Nonstandard Modbus port (default is 502).
+ *   options.localAddress: Local IP address to bind to, default is any.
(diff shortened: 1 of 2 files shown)


yaacov:
Woo, nice !

can you update the option type here:
https://github.com/yaacov/node-modbus-serial/blob/master/ModbusRTU.d.ts#L100
and
https://github.com/yaacov/node-modbus-serial/blob/master/ModbusRTU.d.ts#L104
?

yaacov:
Thanks :heart:

yaacov:
ts types added: https://github.com/yaacov/node-modbus-serial/pull/297

Issue #301, opened 2019-11-07
Timed out when reading coils


devmao:
Modbus v7.7.3
Windows 10 WSL Ubuntu

Just running the `server.js` and the `simple.js` examples I'm getting a time out error when reading the coils.

Output:
```
Reading registers...
Connected
Receive: [ 8005, 8006, 8007, 8008 ]
Reading coils...
Timed out
```

Output with v7.6.0:
```
Reading registers...
Connected
Receive: [ 8005, 8006, 8007, 8008 ]
Reading coils...
Receive: [ false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  true,
  false,
  false,
  false,
  false ]
```

yaacov:
HI, thanks :heart:

a. can you run with debug on ?
b. can you the code that does this ?

`export DEBUG=modbus-serial`
https://github.com/yaacov/node-modbus-serial/blob/master/examples/debug.js

yaacov:
Hi, @devmao :+1: 

Can you test the fix in #302 ?

devmao:
Sure thanks

PR #302, opened 2019-11-10
Server should always return on get coil

yaacov:
Fixes: https://github.com/yaacov/node-modbus-serial/issues/301


devmao:
I will thank you

PR #320, opened 2020-02-10
implemented connectUDP() to fix #306

### LINKED ISSUE #306: Add Modbus UDP
Hello, in some case we need to use Modbus over UDP connection (it work exactly like the Modbus TCP but using UDP.
Is it possible to also have a Modbus UDP function ?
Thanks

sirockin:
I've added tests and implementation for udpport and created the necessary wrapper function connectUDP(). I have done only very basic live testing against a Wago 750-880 plc but it seems to work.

### DIFF
--- a/ModbusRTU.d.ts
+++ b/ModbusRTU.d.ts
@@ -19,6 +19,8 @@ export class ModbusRTU {
   connectRTU(path: string, options: SerialPortOptions): Promise<void>;
   connectTCP(ip: string, options: TcpPortOptions, next: Function): void;
   connectTCP(ip: string, options: TcpPortOptions): Promise<void>;
+  connectUDP(ip: string, options: UdpPortOptions, next: Function): void;
+  connectUDP(ip: string, options: UdpPortOptions): Promise<void>;
   connectTcpRTUBuffered(ip: string, options: TcpRTUPortOptions, next: Function): void;
   connectTcpRTUBuffered(ip: string, options: TcpRTUPortOptions): Promise<void>;
   connectTelnet(ip: string, options: TelnetPortOptions, next: Function): void;
@@ -98,6 +100,12 @@ export interface SerialPortUnixPlatformOptions {
 }
 
 export interface TcpPortOptions {
+  port?: number;
+  localAddress?: string;
+  family?: number;
+}
+
+export interface UdpPortOptions {
   port?: number;
   localAddress?: string;
   family?: number;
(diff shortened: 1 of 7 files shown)


yaacov:
Thanks :heart: 

have no way to test it :-) have fun.

yaacov:
fixes #306

sirockin:
I'll let you know how I get on. A couple of other things:

1.  It would be nice to extract the buffer parse/creation functions in udpport and tcpport somewhere common since they repeat each other
2. I noticed there is a listen() method on the dgram mock which does not exist on the real dgram object. I added bind() since this *does* exist and was necessary for my implementation
3. I'm curious how the C701 port works since it doesn't call bind() on the dgram socket.

yaacov:
>    It would be nice to extract the buffer parse/creation functions in udpport and tcpport somewhere common since they repeat each other

You are welcome to re-factor ( carefully :-) )
 
 >   I noticed there is a listen() method on the dgram mock which does not exist on the real dgram object. I added bind() since this does exist and was necessary for my implementation

?! hmm ... if it needs a missing method how does it work with real dgram object ?

sirockin:
> ?! hmm ... if it needs a missing method how does it work with real dgram object ?
listen() doesn't get called from the C701 implementation (nor does bind()). I think it may have been some copy-paste from the TCP mock. Having said that, I can't see how the C701 implementation works without calling bind().

yaacov:
I do not have a C701 machine to test in real life ..
I need to trust the tests, passing the tests, as they were written when the port was tested with real hardware, is the best thing I can do for now.

Ekristoffe:
sirockin:
  thanks for testing on a WAGO PLC. 
yaacov:
  may I ask what machine use the C701 protocol ?
  also when this UDP could be published on npm ? I need to ask another repo who use your npm to add the udp.

yaacov:
@Ekristoffe hi,

> also when this UDP could be published on npm ? 

Happy new release :balloon: :gift: 
https://github.com/yaacov/node-modbus-serial/releases/tag/v7.8.0

sirockin:
> yaacov: may I ask what machine use the C701 protocol ?

The comment at the head of `c701port.js` reads 'Simulate a modbus-RTU port using C701 UDP-to-Serial bridge' so I presume it's some kind of ethernet/serial converter.

yaacov:
> The comment at the head of c701port.js reads 'Simulate a modbus-RTU port using C701 UDP-to-Serial bridge' so I presume it's some kind of ethernet/serial converter.

Yes, it's a mutli protocol converter, It's a proprietary device usually sold with it's own proprietary client , the code here is reversed engineered ... I want to keep the option to support if :-)

PR #321, opened 2020-02-10
Feature/read device identitification

eeerin:
I was able to start a python server and test some of this readDeviceIdenitifcation code, I did find an issue with my code. I should have an actual machine available soon.


yaacov:
hi, weird, this branch has conflicts, can you take a look ?

eeerin:
Fixed, I forgot to update my forked branch.

index.js
```diff
@@ -395,7 +398,11 @@ ModbusRTU.prototype.open = function(callback) {
                         break;
                     case 43:
                         // read device identification
+<<<<<<< HEAD
```
yaacov: ? are you sure :heart:

eeerin: 😱

yaacov: :-)

eeerin: now I am slightly more. ❤️

yaacov:
Thanks !
will merge when green :+1:

eeerin:
@yaacov I was finally able to test this with an actual machine and it works! Please publish when you can 💚

yaacov:
:+1: :birthday: :cat: 
https://github.com/yaacov/node-modbus-serial/releases/tag/v7.8.0

yadidgarbo:
Hello yaacov, Hello eeerin,

I've been trying to get used to it, but without success.
Could you show me a sample call to read function 43.

I would like to read the device information from a PM5560 meter.
This can be queried via an IP address and a port.
The device is also able to return function 43.

Thanks!

yaacov:
> Could you show me a sample call to read function 43.

@yadidgarbo hi,

I added an example here:
https://github.com/yaacov/node-modbus-serial/blob/master/examples/device-identification.js

have fun :cake:

PR #324, opened 2020-02-24
fix(typescript): fixes #323 an error, when importing the module into …

### LINKED ISSUE #323: Typescript cannot import module
I have the same problem like in this [stack overflow post](https://stackoverflow.com/questions/54093920/telling-typescript-to-to-not-add-default-to-new-class). 

An import of
```
import ModbusRTU from 'modbus-serial';
```
Makes the compiler happy, however the transpiled JS code will not work due to the wrong exports in this libraries's js file.

image

Any suggestions on how to resolve this?

**Edit:** What seems like is missing for this to work is the line
```
module.exports.default = module.exports
```
at the end of your `index.js`

AuspeXeu:
…a typescript project

### DIFF
--- a/index.js
+++ b/index.js
@@ -858,3 +858,4 @@ module.exports.TelnetPort = require("./ports/telnetport");
 module.exports.C701Port = require("./ports/c701port");
 
 module.exports.ServerTCP = require("./servers/servertcp");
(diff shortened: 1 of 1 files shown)


yaacov:
Thanks :green_heart: :+1:

PR #334, opened 2020-04-16
fix: Remove error throw from servertcp_handlers

### LINKED ISSUE #333: question: Modbus Exception 2/3 using ServerTCP
Hi, I would like to try emulate this exceptions with `ModbusRTU.ServerTCP` but If I try to throw an error in vectors methods there is no way to catch it. Also if I return an empty array when the read should fail the error is passed to callback but it's also thrown [here](https://github.com/yaacov/node-modbus-serial/blob/master/servers/servertcp_handler.js#L258) should that throw statement be removed?

How can I manage this? Thanks

robertsLando:
Fixes #333


yaacov:
Thanks :+1:

robertsLando:
Thanks to you @yaacov !

PR #339, opened 2020-06-04
Document the functions for each class

JamesNewton:
The fact that it's writeCoil vs setCoil is not documented /anywhere/ that I could find.  Same for several others.


yaacov:
Thanks :green_heart:

PR #347, opened 2020-06-23
Handle disconnection by Modbus/TCP gateway

sjlongland:
Some Modbus/TCP gateways (e.g. the Advantech EKI-1222-AE that prompted this pull request) will disconnect a client after a period of inactivity.

This causes `modbus-serial` to "silently" mark the connection as closed, which then breaks the application using it because it assumes the port is still open.  (Since it didn't call `.close`, it has no reason to believe otherwise!)

`node-red-modbus-serial` tries to work around this for serial interfaces by accessing `_port` directly, which is a real bad idea, better plan would be for the `modbus-serial` library to emit closure events itself.  This pull request:

1. makes the `ModbusRTU` object an `EventEmitter` so that it can provide such events to library users
2. proxies the `close` event from the port to the outside world so client applications can handle the closure
3. extends the TCP-based port interfaces so they emit `close` when a socket is closed, consistent with the serial versions.
4. also fixes a memory leak which is triggered by calling `.open` on a `ModbusRTU` object repeatedly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,10 @@ require("./utils/buffer_bit")();
 var crc16 = require("./utils/crc16");
 var modbusSerialDebug = require("debug")("modbus-serial");
 
+var util = require("util");
+var events = require("events");
+var EventEmitter = events.EventEmitter || events;
+
 var PORT_NOT_OPEN_MESSAGE = "Port Not Open";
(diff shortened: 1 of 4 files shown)


yaacov:
Thanks !

sjlongland:
No problems. :-)

PR #353, opened 2020-08-18
Buffered RTU FC43 Support

eeerin:
I have tested this with a real device so it definitely works now 😄


yaacov:
Thanks !

eeerin:
@yaacov thanks for the merge! Any chance you could bump the version and republish this library?

yaacov:
Sure :cake: 

https://www.npmjs.com/package/modbus-serial/v/8.0.1

PR #360, opened 2020-09-03
Ensure unitID is integer

Alkarex:
Some clients might call something like `setID("1")` with a string instead of a number.
This then fails due to the strongly-typed comparison in `_onReceive()`
https://github.com/yaacov/node-modbus-serial/blob/0937df9eadb2a08c1ac5c90ceb2ec1017505bfe7/index.js#L370
So this PR casts `setID()` to a number.


yaacov:
Thank you :+1:

Issue #303, opened 2019-11-15
R/W different data type (e.g. floats, int32 ... )


hunter7793:
Hello @yaacov. I'm using `modbus-serial` and it's working well, but when i'm reading data(for example) - i'm receiving only unsigned data type. Can you help me and tell how can i read signed or float data (for example Integer data type which have range -32767...32768).
Thank you

yaacov:
Hi, thanks for the issue :+1:

we have an example here:
https://github.com/yaacov/node-modbus-serial/blob/master/examples/buffer.js

you can get the data buffer [1] of the answer by accessing the `buffer`, then you can parse
the raw data using your home-brew algorithm, or usethe `buffer` module [2] for example:
``` js
 client.readInputRegisters(4, 12)
        .then(function(d) {
            var floatA = d.buffer.readFloatBE(0);
```

[1] https://store.chipkin.com/articles/how-real-floating-point-and-32-bit-data-is-encoded-in-modbus-rtu-messages
[2] https://nodejs.org/api/buffer.html#buffer_buf_readfloatbe_offset

hunter7793:
Thank you very much! i will try it and write you later

hunter7793:
i`ve succesfully reading right data type from the buffer . thank you so much for solution!
But i have one more question. Where should i write my buffer array which contains float or integer data type?

yaacov:
> i`ve succesfully reading right data type from the buffer

:+1: 

> Where should i write my buffer

can you give a code example ?

hunter7793:
var ModbusRTU = require("modbus-serial");
var client = new ModbusRTU();

var startAddress = 0,
      arrayOfMyData = [1], // not a buffer, just array of number types
      buf = Buffer.allocUnsafe(2),
      myBuffer = buf.writeInt16BE(arrayOfMyData[0],0) // i want to write this array

client.writeRegisters(startAddress, arrayOfMyData ,function(err) {
            if (!err) {
                console.log('Somethng wrong'
            } else {
                console.log('Done with writing');
            }

where or how should i write "myBuffer" tag?

yaacov:
> where or how should i write "myBuffer" tag?

the `writeRegisters` method also accept a buffer instead of an array, so you can do:
```
var startAddress = 0,
      arrayOfMyData = [1], // not a buffer, just array of number types
      buf = Buffer.allocUnsafe(2),
[...]

PR #366, opened 2020-10-07
fix: add Buffer input value to writeRegisters signature

vlastoun:
Hello, function writeRegisters also accept buffer according to https://github.com/yaacov/node-modbus-serial/issues/303#issuecomment-554272039

I prepared fix to type definition file.


yaacov:
Hi, thank you for the pull request.

PR #371, opened 2020-10-30
move to github actions and fixed leaked listeners

robertsLando:
I'm not able to add `removeAllListeners()` to `tcprtubufferedport.js`, if I add it there there is a failing test

PR #373, opened 2020-10-30
Remove listeners on close tcp buffered

yaacov:
@robertsLando  can you check if this solves the issue in #371 ?


robertsLando:
@yaacov Yes it does!

robertsLando:
@yaacov I'm wondering if it would be correct to also remove listeners from `this._client`. What do you think?

yaacov:
> @yaacov I'm wondering if it would be correct to also remove listeners from this._client. What do you think? On close event I mean

IMHO no, unless it creates a problem for someone :-)

robertsLando:
:+1:

PR #399, opened 2021-03-10
modbus worker

yarosdev:
How to use

```

const client = new ModbusRTU()

client.connect..........

client.setWorkerOptions({
    maxConcurrentRequests: 10, // it will send 10 requests or less at a time if any
    debug: true
})

// Read 4 values starting from 10009, under the hood it will read 8 registers due to int32 type
const response1 = await client.send({
        unit: 1,
        fc: 3,
        address: 10009,
        quantity: 4,
        type: 'int32',
    });
// Write 2 values to address: 10009 and 10011
const response2 = await client.send({
        unit: 1,
        fc: 16,
        address: 10009,
        value: [10999, 10888],
        type: 'int32',
    });

// Start poll
// under the hood it will build optimal requests for you 
client.poll({
        unit: 1,
        map: [
            { fc: 3, address: [10011, 10013, 10018], type: "int32" },
            { fc: 3, address: 10003, type: "int32" },
            { fc: 3, address: 10005, type: "int32" },
            { fc: 3, address: 10007, type: "int32" },
            { fc: 3, address: 10009, type: "int32" },
            { fc: 2, address: [1,2,3]},
            { fc: 1, address: [1,2,3]},
            { fc: 1, address: 4},
            { fc: 1, address: 5},
            { fc: 1, address: 6},
            { fc: 3, address: [10001]}, // default type is int16
            { fc: 3, address: [10020, 10023, 10026], type: "float"},
            { fc: 3, address: [10030, 10034], type: "double"}
        ],
        onProgress: (progress, data) => {
            console.log("poll #1 - ", progress.toFixed(2)+"%", JSON.stringify(data))
        },
        maxChunkSize: 32, // max registers per request
        skipErrors: false, // if false it will stop poll scenario and return partial result
    })
        .then(res => console.log('poll #1', res))
        .catch(res => console.log('poll #1 with errors', res))

```

requests made by command "send" has **greater** priority then "poll", it means your poll requests will be executed right after normal requests. It is for situations when you polling hundreds of registers, but same time you are able to write some registers as soon as possible (after first poll requests completed)

available types are **int16**, **uint16**, **int32**, **uint32**, **float**, **double**

### DIFF
--- a/apis/connection.js
+++ b/apis/connection.js
@@ -35,7 +35,7 @@ var addConnctionAPI = function(Modbus) {
             obj.open(next);
         } else {
             // o/w use  a promise
-            var promise = new Promise(function(resolve, reject) {
+            return new Promise(function(resolve, reject) {
                 function cb(err) {
                     if (err) {
                         reject(err);
@@ -46,8 +46,6 @@ var addConnctionAPI = function(Modbus) {
 
                 obj.open(cb);
             });
-
-            return promise;
         }
     };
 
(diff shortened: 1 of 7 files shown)


yaacov:
Nice +1
can you add a README.ms in the worker dir, and add some words an a link in the main README ?

yaacov:
ahhh, and example script in the example dir ?

yarosdev:
OK I`ll add later

yaacov:
Thanks :green_heart:

Issue #415, opened 2021-06-22
When returning error code from getMultipleHoldingRegisters, shouldn't need to return values array 


bruceceng:
Consider the following code. If the address is out of range, then it should return error code 0x02. However, when using the callback, you must added the `,values` otherwise it will return a different error code (0x04).

```    
    getMultipleHoldingRegisters: function(addr, length, unitID, callback) {
        console.log("getMultipleHoldingRegisters", addr, length);

        let values = Array(length).fill(0); //for some reason even when returning an error we need to set a values array.
        //check the address
        const minAddr = discreteInputCount + writeReportingCoilCount + automaticCoilCount + inputRegisterCount;
        const maxAddr = minAddr + writeReportingHoldingRegisterCount + automaticHoldingRegisterCount - 1;

        if (addr < minAddr || addr + length - 1 > maxAddr) {
            console.log("illegal address");
            callback({
                modbusErrorCode: 0x02, // Illegal address
                msg: "Invalid address"
            },values);
            return;
        }

        const err = null;
        
        for (let i=0; i<length; i++) {
            const dataIndex = addr - minAddr + i;
            values[i] = holdingRegisterValues[dataIndex];
        }
        callback(err,values);
    },
````

The problem is here in servertcp_handler: 

```
                if (!err && values.length !== length) {
                    var error = new Error("Requested address length and response length do not match");
                    callback(error);
                } else {
                    for (var i = 0; i < length; i++) {
                        var cb = buildCb(i);
                        try {
                            cb(err, values[i]);
                        }
                        catch (ex) {
                            cb(ex);
                        }
                    }
                }
            });
```

There should be another `else if` to detect the return of an error rather than evaluating values[i] which triggers the catch or at minimum this should be documented. Right now, the callback from getHoldingRegister doesn't require a value to be passed if an error is returned, but the callback form getMultipleHoldingRegisters does.

PR #416, opened 2021-06-22
Improve server callbacks

bruceceng:
Fixed issue #415 - [When returning error code from getMultipleHoldingRegisters, shouldn't need to pass values array to callback](https://github.com/yaacov/node-modbus-serial/issues/415 )

Also when _handleWriteMultipleRegisters was called, it checked for the presence of setRegister before setRegisterArray so it was impossible to build seperate handlers for both. This was corrected by changing the order.

Also there was an issue with the number of parameters a handler with a callback should have for setRegisterArray (6 vs 4)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -80,7 +80,7 @@
         "keyword-spacing": "off",
         "line-comment-position": "off",
         "linebreak-style": [
-            "error",
+            "off",
             "unix"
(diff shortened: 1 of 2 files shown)


yaacov:
@bruceceng hi, thank you for the fix, can you address the failing tests ?

PR #439, opened 2022-01-06
Add timeout option to TcpPortOptions

dyllan-to-yu:
It looks like `timeout` is a supported option; this adds it to the TS definition


yaacov:
Thanks :green_heart:

PR #450, opened 2022-04-03
Implement Bluetooth LE port

Turbo87:
This implements a ModBus port based on Bluetooth Low Energy. A TX service and characteristic is used for writing data and a RX service/characteristic is used for receiving the responses.

This kind of communication is used by devices like https://www.renogy.com/bt-2-bluetooth-module/ that essentially act as a ModBus to Bluetooth LE bridge.


yaacov:
wow, thanks :green_heart:

PR #451, opened 2022-04-05
Convert to native class syntax

Turbo87:
All [currently maintained Node.js versions](https://nodejs.org/en/about/releases/) support the native class syntax these days. This PR converts all of the `prototype` based classes to native classes to make the code a little easier on the eyes. :)

Review with "Hide whitespace" is recommended: https://github.com/yaacov/node-modbus-serial/pull/451/files?w=1 😅


yaacov:
Thanks :green_heart: :+1:

yaacov:
> Review with "Hide whitespace" is recommended

:smile:

PR #453, opened 2022-04-06
Use let/const instead of var

Turbo87:
Support for `let` and `const` has existed in Node.js since v6.0.0 and the minimum Node.js version that this package appears to support currently is Node.js v10. In other words, it should be safe to use these new keywords which essentially replace the `var` keyword.

This PR also enables the `no-var` ESLint rule, which prevents `var` usage from now on. The only exception is the `servers/servertcp_handler.js` file, where `var` is currently used in questionable ways which can't directly translate to let/const.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -200,7 +200,7 @@
         "no-useless-escape": "error",
         "no-useless-rename": "error",
         "no-useless-return": "off",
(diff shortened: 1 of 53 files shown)


yaacov:
Wow :exclamation: 
:green_heart:

PR #458, opened 2022-05-25
Remove removeAllListeners from close callback.

glengiffey:
In index.js line 523, "modbus._port.once("close", modbus.emit.bind(modbus, "close"));" is not called when we use the close method on line 165.
https://github.com/yaacov/node-modbus-serial/blob/master/index.js#L523

I believe the reason for this is that the method calls client.end() and then removeAllListeners(). What actually should happen is that we should wait until we emit close. Which we do on line 110.
https://github.com/yaacov/node-modbus-serial/blob/master/ports/tcpport.js#L110

So I think this removeAllListeners is redundant and causes undesirable behavior.


yaacov:
:+1: thanks

PR #464, opened 2022-06-29
fix: SocketError -> socketError

everhardt:
in the rest of the project `socketError` (starting with a lower case `s`) is used


yaacov:
:green_heart:

PR #465, opened 2022-06-29
feat: on("close") in ModbusRTU.d.ts

everhardt:
Add the `close` event that [ModbusRTU emits](https://github.com/everhardt/node-modbus-serial/blob/bdffd759a279f959a7e250f0f1858bc6ad569607/index.js#L523-L525) to the Typescript interface.


yaacov:
:green_heart:

PR #479, opened 2022-11-16
feat:Add ModbusRTU TestPort interface

chaos2171053:
Should add ModbusRTU.TestPort interface in typescript development mode.


yaacov:
Thanks 👍

Issue #482, opened 2022-11-25
fail on tsc buid with version 8.0.7


dbettoni:
On last version 8.0.7 my tsc build produce this error:
> tsc

node_modules/modbus-serial/ModbusRTU.d.ts:2:22 - error TS7016: Could not find a declaration file for module './ports/testport'. '...../node_modules/modbus-serial/ports/testport.js' implicitly has an 'any' type.

2 import TestPort from './ports/testport';

8.0.6 is OK

yaacov:
Hi, thanks for the issue :fire:

Can you make a pull request fixing it ?

dbettoni:
I can try but I don't know when or time to do it,
for now the project works by setting the dependency to "modbus-serial": "8.0.6",

yaacov:
The issue looks related to https://github.com/yaacov/node-modbus-serial/pull/479
@chaos2171053 hi, can you take a look ?

@dbettoni can you add more information that will help understand the problem, link to your code ?
for example:
how you use the module in your project ?
how do you get the error ?

chaos2171053:
> On last version 8.0.7 my tsc build produce this error:
> 
> > tsc
> 
> node_modules/modbus-serial/ModbusRTU.d.ts:2:22 - error TS7016: Could not find a declaration file for module './ports/testport'. '...../node_modules/modbus-serial/ports/testport.js' implicitly has an 'any' type.
> 
> 2 import TestPort from './ports/testport';
> 
> 8.0.6 is OK

fix on this [pr](https://github.com/yaacov/node-modbus-serial/pull/484#issue-1467366277) .Can u provie more information to repoduce this bug?May be i can make this pr better.

yaacov:
#484 is merged @dbettoni can you check the fix ?

dbettoni:
I tried to rebuild my project by importing modbus-serial 8.0.7 but the problem hasn't been resolved.
see attached image
<img width="2064" alt="Schermata 2022-11-29 alle 12 44 47" src="https://user-images.githubusercontent.com/30083487/204521922-09b08323-e5c1-4757-8da0-39b8618278b1.png">

dbettoni:
my tsconfig.json:
<img width="376" alt="Schermata 2022-11-29 alle 14 14 22" src="https://user-images.githubusercontent.com/30083487/204538782-25ba7087-bb0a-482f-b803-82e18c8d61cf.png">

yaacov:
[...]

PR #486, opened 2022-12-01
Update TestPort.d.ts

mchakroun:
bug fix for https://github.com/yaacov/node-modbus-serial/issues/482#issuecomment-1333740361


yaacov:
@mchakroun nice, can you add a test to make sure we don't brake the types in the future ?

mchakroun:
> @mchakroun nice, can you add a test to make sure we don't brake the types in the future ?
@yaacov https://github.com/yaacov/node-modbus-serial/pull/487 this should do it

Issue #488, opened 2023-01-16
How Create ModbusRTUSlave?


teddy1565:
Hi, i tried create ModbusRTUSlave.
but the module looks like it's still in the works.

No available serialserver can use. it seems to share the servertcp_handler.js
I think there is something wrong with some interfaces.

for example:

An unknow error, blocking serialport:

ServerSerial Constructor:

```javascript
options = options || {};

const optionsWithBinding = {
    path: options.port || PORT,
    baudRate: options.baudrate || BAUDRATE,
    debug: options.debug || false,
    unitID: options.unitID || 255
};
if (options.binding) optionsWithBinding.binding = options.binding;

// create a serial server
modbus._server = new SerialPort(optionsWithBinding);
```
Make the following changes:

```javascript
 modbus._server = new SerialPort({
      path: options?.path ?? PORT,
      baudRate: options?.baudRate ?? BAUDRATE,
      debug: options?.debug ?? false,
      unitID: options?.unitID ?? 255
});
```

I think it would be better to synchronize the option with the key of the SerialPort interface. 
Reduce misunderstandings under missing files.

another problem is serialport recive buffer with time difference.
so when i try get buffer, console tell me, i get this:
```
<Buffer 01 05 00 05 ff 00>
<Buffer 9c 3b>
```
but we expect can get:
```
<Buffer 01 05 00 05 ff 00 9c 3b>
```

so i thinik, we can use @serialport/parser-inter-byte-timeout, or other parser package, 
ensure parse only when we receive all packets.

Finally, I would like to ask whether it is possible to export serial server interface,
and have them referenced differently in the development environment
Because it unreasonable to reference serverTCP_handler in serverSerial

English is not my mother tongue, sorry for any offense

yaacov:
hi, thank you for your question, I added a "help wanted" label

you can try to look at the examples, you may find an answer in there:
https://github.com/yaacov/node-modbus-serial/tree/master/examples

teddy1565:
thanks, but examples doesn't have for serverSerial.
```javascript
const ModbusRTU = require("modbus-serial")
const serverRTU = new ModbusRTU.serverSerial(options)
```
The implementation of serverSerial and serverTCP should be different.
i think just solve the serialPort Parser & export interface.
the serverSerial can works great

teddy1565:
It can indeed pass in the CI test, but may need to consider physical factors.
[...]

PR #489, opened 2023-01-17
ModbusRTU Slave (Server Serial)

teddy1565:
#488 

## ModbusRTU Slave (Server Serial) fix
- a serialport timeout-parser for SerialServer
- create server options interface sync to SerialPort
  ```javascript
   const options = {
       path: "/dev/ttyUSB0",
       baudRate: 9600,
       debug: false,
       unitID: 1
   }
   const RTUSlave = new ModbusRTU.ServerSerial(vector, options);
  ```
- make responseBuffer and Close port use different of  ParserPort (Otherwise it will cause echo, that will looks like blocking)

Because in CI test It cannot reflect the real hardware and the packet receiving and sending status when passing through modbusRTU.
If just use modbusTCP Handler, When serialport encounters a data packet and the receiving time is a little longer, an unparseable situation will occur.
So I wrote with reference to the timeout-parser in SerialPort. The reason for not using serialPort's parser directly is to reduce third-party dependencies.
Allows developers to adjust the delay parameters of receiving packets according to the actual device packet delay

So I add a timeout-parser for SerialServer and I try to get parameter option to sync with SerialPort. 
Avoid misunderstandings in the absence of documents and instructions
but I'm still compatible with the past, like:
```javascript
optionswithbinding = {
path: options.path || oprtions.port || PORT,
}
```

Finally, it was found that calling the correct port of responseBuffer will cause unexpected recursive blocking, so in the two events of responseBuffer and close, the original path is used instead of the path after the parser

## Original description:

- 增加了一個適用於serialport的timeout-parser
- 令modbus.ServerSerial 建構子的options 與 SerialPort建構子的options同步，避免採用與TCP相同的參數導致誤會
- 修正了使用parser時，由於監聽port與寫入的port同樣而導致意外無止盡遞迴(那將會看起來像是阻塞)

由於原先的CI測試中，並沒有考量到物理層面的延遲，因此無法正確地接收Packet解析。
所以我參考了SerialPort中的timeout-parser進行撰寫，不直接使用serialPort的解析器是為了降低第三方依賴。
讓開發者能夠根據實際裝置封包延遲情況，調整接收封包的延遲參數

並嘗試讓modbus.ServerSerial的建構子參數與SerialPort的建構子參數相似，減少參考資料及文件缺失的情況下的誤會。

最後發現responseBuffer調用部正確的port會造成意外的遞迴阻塞，因此responseBuffer以及close這兩個事件中，使用了原始的path而非經過解析器後的path

### DIFF
--- a/servers/serverserial.js
+++ b/servers/serverserial.js
@@ -18,6 +18,7 @@ const events = require("events");
 const EventEmitter = events.EventEmitter || events;
 const modbusSerialDebug = require("debug")("modbus-serial");
 const { SerialPort } = require("serialport");
+const ServerSerialPipeHandler = require("./serverserial_pipe_handler")
 
 const PORT = "/dev/tty";
 const BAUDRATE = 9600;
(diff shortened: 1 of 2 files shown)


yaacov:
Thanks :green_heart:

PR #491, opened 2023-02-03
fix serialport server bug

teddy1565:
- getPort method will return Serialport instance not parser instance
- when create serialport, pass serialport parameter to serialport constructor, not just ignore. Contributes to developer flexibility when using the suite


yaacov:
thanks :green_apple:

PR #493, opened 2023-03-03
Enron

1038642892:
This PR is to allow the TCP server (starting with this server first) to handle 32 bit ENRON registers.

Taken specification from here:
https://www.simplymodbus.ca/Enron.htm


index.js
```diff
@@ -675,11 +719,15 @@ class ModbusRTU extends EventEmitter {
         // function code defaults to 4
         code = code || 4;
 
+        let valueSize = 2;
+        if (this._enron && !(dataAddress >= this._enronTables.shortRange[0] && dataAddress <= this._enronTables.shortRange[1])) valueSize = 4;
+
```
yaacov, line 724:
```suggestion
        // value size for Modicon modbus is 2
        let valueSize = 2;
        // value size for Enron Modbus is 4
        if (this._enron && !(dataAddress >= this._enronTables.shortRange[0] && dataAddress <= this._enronTables.shortRange[1])) {
              valueSize = 4;
        }
```

yaacov:
:green_heart: thanks

yarosdev:
Long time ago I have pushed here separate file called "worker" which can read int32

yarosdev:
Why did you implement "ranges"? Why not just call needed method? read/writeEnronRegister ? read/writeEnronBool? Etc.
Ranges are not needed here at all, because some device vendors just not follow all rules of protocol specification. 

I have used a device, that have same register address in every table because it under the hood - transforms a register address into a physical memory address.... 

I have used a device that have few addresses that do not support read fc at all - only write function.

And so on..

Have you asked yourself, "why this library has method for each function of modbus protocol?"?

You can only implement each function of Enron modbus, but not implement what you don't know how it is in real life in real devices...

yarosdev:
IMG_20230305_000357

As you can see, a table is being selected by a function code, not a register address. Read specification more careful before implementation.

yarosdev:
And it's better to create a separate class ModbusEnron because it is a different protocol... It is strange to have "read holding register" along side with "read Enron register"

Refactoring, refactoring....

yarosdev:
And one more, numeric variables should also work with buffers as values... Cause you don't know how device encode a data value... at least BE or LE

yaacov:
@yarosdev thanks, can you make a pull request with your suggestions ?

yarosdev:
I can, for know I am not sure about my free time, I will keep this in mind for a while

1038642892:
@yarosdev Thank you for your comments, I'm not following your explanations very well, could we focus on one point at a time?

Let's go with ranges first.

You said "Ranges are not needed here at all, because some device vendors just not follow all rules of protocol specification."

Ranges allow you to define them in configuration rather than hard coded in the library - hence should allow for flexibility when specification isn't followed by vendors.

yarosdev:
@1038642892  There can be a case when ranges overlapping

PR #495, opened 2023-03-14
add FC17 support (report server ID)

peuter:
This function code is used to read the description of the type, the current status, and other
information specific to a remote device.


yaacov:
Thanks :+1:

PR #498, opened 2023-04-11
Add `getPorts` method

robertsLando:
Most projects using this module will also have to install serialport in order to only list ports, this could lead having multiple serialport versions installed with some issues expecially when cross compiling. With this util method we expose `Serialport.list` to prevent this


yaacov:
:green_heart:

robertsLando:
Thanks for the quick follow up @yaacov 🙏🏼  Make a release when possible, thanks 💚

robertsLando:
thanks 🚀

PR #499, opened 2023-04-12
Update IServiceVector deffinitions

HighWolf:
- getMultipleInputRegisters and getMultipleHoldingRegisters functions must return an array of numbers since those functions process multiple words.
- setRegisterArray function is also working and ready to be implemented.


yaacov:
Thanks :green_heart:

PR #507, opened 2023-08-17
Rtu socket

1038642892:
We have TCP (MBAP header) but no RTU over Socket (TCP) server. Implemented it in this PR.


yaacov:
Nice, thanks !

PR #510, opened 2023-08-29
Bump serial port to 12.0.0

yaacov:
Bumping dependencies versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 18.x]
+        node-version: [14.x, 18.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


yaacov:
@Jalle19 hi, can you check if this version bumps change something for your use case ?

yaacov:
Closing,
seems that reverting #507 fixed the issue.

Jalle19:
Did you force push away the changes or what happened?

yaacov:
> Did you force push away the changes or what happened?

yes, I reverted #507 and #508 and forced push it.

1038642892:
Can I get a bit more detail please @Jalle19 as to what broke?

Jalle19:
After reading a couple of registers, all Modbus requests started timing out. I don't have an easy setup for debugging Modbus (no test devices), so I can't really give more information.

Jalle19:
https://github.com/Jalle19/eda-modbus-bridge/blob/master/eda-modbus-bridge.mjs#L91 
https://github.com/Jalle19/eda-modbus-bridge/blob/master/app/modbus.mjs

1038642892:
@Jalle19 thanks, which server/device set up are you using? Looking at the debug log, is there only one query that times out? Do the others return OK?

PR #512, opened 2023-08-29
Bump dependencies versions

yaacov:
Bump some dependencies

Noteed change:
serialport is now 12


yaacov:
@jalle19 does your use case still work with this changes (e.g. serial port 12) ?

PR #519, opened 2023-10-18
Allow to pass socket options to TCP client

AndreMaz:
Hi @yaacov thank you for the lib. It's been super useful

While doing some debugging I've made some changes to the lib and decided to create a PR.

Here are the changes:
1) allows to pass socket opts (link to docs: https://nodejs.org/api/net.html#new-netsocketoptions)  to the TCP client. When creating `new Socket()`
2) allows to pass TCP connections (link to docs: https://nodejs.org/api/net.html#socketconnectoptions-connectlistener) when calling `_client.connect()`
3) adds a test coverage reporter that helps to see uncovered codebase
image
4) adds a vscode launch.json that allows to attach a debugger to unit tests and to the examples in `examples` dir
5) adds and example showing how to use 1) to pass abort signal to the Socket() and easily close the socket.

Re 5) I have a question for you. When the socket emits an error and socket connection is closed, does it make sense to wait for a response (which won't come as socket is closed) and then throw a `TransactionTimedOutError` ?
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,3 @@ modbus-serial
 docs
 
 package-lock\.json
-
-# Visual Studio Code
-# .vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+.nyc_output
 
 # Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
 .grunt
@@ -35,4 +36,4 @@ docs
 package-lock\.json
 
 # Visual Studio Code
-.vscode/
+# .vscode/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
         "es6": true
     },
     "parserOptions": {
-        "ecmaVersion": 2017
+        "ecmaVersion": 2020
     },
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -35,4 +36,4 @@ docs
 package-lock\.json
 
 # Visual Studio Code
-.vscode/
+# .vscode/
```
yaacov, line 39: if you think it's useful, you can remove it

AndreMaz: Yep. I think that vast majority of JS devs use vscode so it's useful to have a debugger config.

Action: Removed the `vscode/` from gitignore

AndreMaz:
It fails for node v14 because AbortController was introduced in node v15 (source: https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility) . However, node 14 reached EOL at April 30, 2023

yaacov:
> It fails for node v14 because AbortController was introduced in node v15 (source: https://developer.mozilla.org/en-US/docs/Web/API/AbortController#browser_compatibility) . However, node 14 reached EOL at April 30, 2023

I am ok with removing v14 from the test matrix

yaacov:
Thank you for this pull request :green_heart:

AndreMaz:
@yaacov dropped node v14. Thank you for reviewing the code.

PR #521, opened 2023-10-23
Cancel pending transactions when calling destroy

AndreMaz:
This PR cancels pending transactions (under the hood calls `clearTimeout()` ) when the `destroy()` method is called.

Rationale, after `destroy()` is called it does not make sense to wait for a response hence we need to cancel all pending requests. 

How to test:
- On modbus-sever, add a delay of 10secs on `getHoldingRegister()`
- Set client timeout to 7 secs
- Call `readHoldingRegisters()` in the example below
- after 5 secs try to destroy the client

Repro example
```js
"use strict";

const ModbusRTU = require("../index");

async function main() {
  try {
    await client.connectTCP("127.0.0.1", { port: 8502 });
    console.log("Connected");
    client.setID(1);
    client.setTimeout(7_000);

    // Disconnect after 5 seconds
    setTimeout(async () => {
      try {
        console.log(`Destroying connection at ${new Date().toISOString()}`);
      await new Promise((resolve, reject) => {
        // Try to close the connection
        client.destroy((err) => {
          if (err) return reject(err);

          console.log(`Destroyed closed at ${new Date().toISOString()}`);
          return resolve();
        });
      });
      } catch (error) {
        console.log(`Failed to destroy connection: ${error.message} at ${new Date().toISOString()}`);
      }
      
    }, 5000);

    await readRegistersOneByOne();
  } catch (error) {
    console.log(error.message);
  } finally {
    client.close();
  }
}

const spec = {
  telemetry: {
    regA: {
      address: 1,
      registerCnt: 1,
    },
  },
};

async function readRegistersOneByOne() {
  for (const [key, { address, registerCnt }] of Object.entries(
    spec.telemetry
  )) {
    try {
      console.log(`Reading registers at ${address} with ${registerCnt} registers`);
      const data = await client.readHoldingRegisters(address, registerCnt);
      console.log(key, data.buffer);
    } catch (error) {
      console.log(error);
    }
  }
}

main();
```

Without "cleanup"
image

With "cleanup"
image

### DIFF
--- a/index.js
+++ b/index.js
@@ -638,6 +638,23 @@ class ModbusRTU extends EventEmitter {
     }
 
     /**
+     * Clears the timeout for all pending transactions.
+     * This essentially cancels all pending requests.
+     */
+    _cancelPendingTransactions() {
+        if (Object.keys(this._transactions).length > 0) {
+            Object.values(this._transactions).forEach((transaction) => {
+                if (transaction._timeoutHandle) {
+                    _cancelTimeout(transaction._timeoutHandle);
+                }
+            });
+        }
+    }
+
+    // eslint command to fix this file
+    // eslint --fix index.js
+
+    /**
      * Close the serial port
      *
      * @param {Function} callback the function to call next on close success
@@ -661,6 +678,9 @@ class ModbusRTU extends EventEmitter {
      *      or failure.
      */
     destroy(callback) {
+        // cancel all pending requests as we're closing the port
+        this._cancelPendingTransactions();
+
         // close the serial port if exist and it has a destroy function
         if (this._port && this._port.destroy) {
             this._port.removeAllListeners("data");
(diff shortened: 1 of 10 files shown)


index.js
```diff
@@ -637,6 +637,23 @@ class ModbusRTU extends EventEmitter {
...
+            Object.values(this._transactions).forEach((transaction) => {
+                if (transaction._timeoutHandle) {
+                    _cancelTimeout(transaction._timeoutHandle);
+                }
+            });
+        }
+    }
+
+    // eslint command to fix this file
+    // eslint --fix index.js
```
yaacov, line 655: hi, thank you for the pull request, do we need this lines ?

AndreMaz: Sorry about that. It was a self note to format the code before creating a commit. 

But I guess I forgot to delete it

PR #533, opened 2023-11-28
Fix readme

yarosdev:
Added info about modbusdb and example. Example in the readme is a working example)


worker/README.md
```diff
@@ -1,3 +1,8 @@
+# DEPRECATED
```
yaacov, line 1: can you change to someting like 'for a more complete module see modbusdb ....

depracated is not correct here, using modbusdb is another option, it has some benefits over the worker, but it is also another dependency some users will opt to avoid

yarosdev: changed

yaacov:
Thank you for the pull request

Issue #536, opened 2023-12-12
ServerTCP causes uncaught exception if listen fails


everhardt:
Take this example with a host it cannot bind to (`127.0.0.1`):
```ts
import { ServerTCP } from 'modbus-serial';

async function go(): Promise<void> {
  const server = new ServerTCP(
    {},
    {
      host: '127.0.0.1',
      port: 502,
    },
  );
  await new Promise<void>((resolve) => {
    server.on('initialized', () => {
      console.log('initalized');
      resolve();
    });
  });
}

process.on('uncaughtException', (error) => {
  console.log('uncaughtException');
  console.error(error);
  process.exit(1);
});

go()
  .then(() => {
    process.exit(0);
  })
  .catch((e) => {
    console.log('caught exception');
    console.error(e);
    process.exit(1);
  });

```

running it results in the following console output:
```
uncaughtException
Error: listen EACCES: permission denied 127.0.0.1:502
    at Server.setupListenHandle [as _listen2] (node:net:1800:21)
    at listenInCluster (node:net:1865:12)
    at doListen (node:net:2014:7)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'EACCES',
  errno: -13,
  syscall: 'listen',
  address: '127.0.0.1',
  port: 502
}
```

PR #537, opened 2023-12-12
fix: ServerTCP causes uncaught exception if listen fails

everhardt:
fixes https://github.com/yaacov/node-modbus-serial/issues/536

I've opted for a separate event (`serverError`), as `error` is already used and the with the two kinds will probably have to be dealt with differently.


yaacov:
Thank you for the pull request

PR #543, opened 2024-01-31
fix: catch error thrown in getMultipleHoldingRegisters

everhardt:
It's inconsistent that a rejected promise returned by getMultipleHoldingRegisters and a callback with and error are handled, but a synchronous throw is not. This PR fixes that.


yaacov:
:green_heart: Thanks

PR #545, opened 2024-02-08
Feature: Expand error codes

sjlongland:
A lot of the Modbus exception codes are not defined, leading to `Unknown error` as the response for things like gateway timeouts and such.  This defines all the common Modbus exception codes defined in the Modbus specification.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,18 +35,18 @@ const TRANSACTION_TIMED_OUT_MESSAGE = "Timed out";
 const TRANSACTION_TIMED_OUT_ERRNO = "ETIMEDOUT";
 
 const modbusErrorMessages = [
-    /*code 0=*/"Unknown error",
-    /*code 1=*/"Illegal function (device does not support this read/write function)",
-    /*code 2=*/"Illegal data address (register not supported by device)",
-    /*code 3=*/"Illegal data value (value cannot be written to this register)",
-    /*code 4=*/"Slave device failure (device reports internal error)",
-    /*code 5=*/"Acknowledge (requested data will be available later)",
-    /*code 6=*/"Slave device busy (retry request again later)",
-    /*code 7=*/"Negative acknowledge (slave device cannot perform programming functions)",
-    /*code 8=*/"Memory parity error (slave device detected a parity error in memory)",
-    /*code 9=*/"Unknown error",
-    /*code10=*/"Gateway path unavailable (misconfigured gateway)",
-    /*code11=*/"Gateway target device failed to respond (retry request again later)"
+    "Unknown error",
+    "Illegal function (device does not support this read/write function)",
+    "Illegal data address (register not supported by device)",
+    "Illegal data value (value cannot be written to this register)",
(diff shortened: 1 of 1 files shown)


yaacov:
wow, thank you for the pull request

Ref:
| Code (Hex) | Name                                  | Meaning                                                                                                                                                                                                                                           |
|------------|---------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 01         | Illegal Function                      | The function code received in the query is not an allowable action for the slave. This may be because the function code is only applicable to newer devices, and was not implemented in the unit selected. It could also indicate that the slave is in the wrong state to process a request of this type, for example because it is unconfigured and is being asked to return register values. If a Poll Program Complete command was issued, this code indicates that no program function preceded it. |
| 02         | Illegal Data Address                  | The data address received in the query is not an allowable address for the slave. More specifically, the combination of reference number and transfer length is invalid. For a controller with 100 registers, a request with offset 96 and length 4 would succeed, a request with offset 96 and length 5 will generate exception 02.                                                                                       |
| 03         | Illegal Data Value                    | A value contained in the query data field is not an allowable value for the slave. This indicates a fault in the structure of the remainder of a complex request, such as that the implied length is incorrect. It specifically does NOT mean that a data item submitted for storage in a register has a value outside the expectation of the application program, since the MODBUS protocol is unaware of the significance of any particular value of any particular register.              |
| 04         | Slave Device Failure                  | An unrecoverable error occurred while the slave was attempting to perform the requested action.                                                                                                                                                                                                             |
| 05         | Acknowledge                           | Specialized use in conjunction with programming commands. The slave has accepted the request and is processing it, but a long duration of time will be required to do so. This response is returned to prevent a timeout error from occurring in the master. The master can next issue a Poll Program Complete message to determine if processing is completed.                                 |
| 06         | Slave Device Busy                     | Specialized use in conjunction with programming commands. The slave is engaged in processing a long-duration program command. The master should retransmit the message later when the slave is free.                                                                                                         |
| 07         | Negative Acknowledge                  | The slave cannot perform the program function received in the query. This code is returned for an unsuccessful programming request using function code 13 or 14 decimal. The master should request diagnostic or error information from the slave.                                                            |
| 08         | Memory Parity Error                   | Specialized use in conjunction with function codes 20 and 21 and reference type 6, to indicate that the extended file area failed to pass a consistency check. The slave attempted to read extended memory or record file, but detected a parity error in memory. The master can retry the request, but service may be required on the slave device.                   |
| 0A         | Gateway Path Unavailable              | Specialized use in conjunction with gateways, indicates that the gateway was unable to allocate an internal communication path from the input port to the output port for processing the request. Usually means the gateway is misconfigured or overloaded.                                                  |
| 0B         | Gateway Target Device Failed to Respond | Specialized use in conjunction with gateways, indicates that no response was obtained from the target device. Usually means that the device is not present on the network.                                                                                                                                  |

Issue #503, opened 2023-06-29
ServerSerial usage?


MillerRen:
Is this lib can as a serial server? I can see a unit test but not found the usage.

1038642892:
Yes check [this example](https://github.com/yaacov/node-modbus-serial/blob/master/examples/server_enron_serial.js) please, but you're right the documentation is lacking, will add some.

PR #550, opened 2024-04-04
ServerSerial types and usage

cplepage:
I was looking on how to use the ServerSerial in my TypeScript project, but had to look in the code to figure it out.

This is quickly patched to be able to work with TypeScript, but I'm starting the conversation to make it to the codebase.

Open to comment, feedback and reviews.

Related to #503

### DIFF
--- a/servers/serverserial.js
+++ b/servers/serverserial.js
@@ -237,7 +237,7 @@ class ServerSerial extends EventEmitter {
         const optionsWithBindingandSerialport = Object.assign({}, serialportOptions, optionsWithBinding);
 
         // create a serial server
-        modbus._serverPath = new SerialPort(optionsWithBindingandSerialport);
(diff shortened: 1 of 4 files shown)


yaacov:
Hi, thank you for the pull request

servers/serverserial.js
```diff
@@ -253,7 +253,7 @@ class ServerSerial extends EventEmitter {
         // remember open sockets
         modbus.socks = new Map();
 
-        modbus._server.on("open", function() {
```
yaacov, line 256: Hi, can you give some description what this do, and why this change is needed ?

cplepage: The `open` event never gets emitted.
I think it may be because of how `ServerSerialPipeHandler` works? 
But the `open` event from `SerialPort`  isn't piped and so you never receive an `initialized` event either.

PR #557, opened 2024-06-07
fix: uncaught error in read device identification

everhardt:
I ran into the following error when communicating with a solar logger over Modbus TCP:
```
RangeError [ERR_OUT_OF_RANGE]: The value of "offset" is out of range. It must be >= 0 and <= 12. Received 43
     at __node_internal_captureLargerStackTrace (node:internal/errors:490:5)
     at new NodeError (node:internal/errors:399:5)
     at boundsError (node:internal/buffer:88:9)
     at Buffer.readUInt8 (node:internal/buffer:254:5)
     at _readFC43 (/app/node_modules/modbus-serial/index.js:267:40)
     at ModbusRTU._onReceive (/app/node_modules/modbus-serial/index.js:538:13)
     at TcpPort.emit (node:events:513:28)
     at Socket.<anonymous> (/app/node_modules/modbus-serial/ports/tcpport.js:122:22)
     at Socket.emit (node:events:513:28)
     at addChunk (node:internal/streams/readable:324:12)
     at readableAddChunk (node:internal/streams/readable:297:9)
     at Readable.push (node:internal/streams/readable:234:10)
     at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

Two problems here:
1. the error was uncaught, ie. it crashed my program
2. there are two interpretations of `numOfObjects`: the number of objects in the response or the number of objects in all responses combined. The [spec](https://modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf) states the former, but  the example on page 45 shows the latter. The solar logger I was communicating with apparently uses the latter as well.

This PR addresses both problems.

### DIFF
--- a/index.js
+++ b/index.js
@@ -268,7 +268,10 @@ function _readFC43(data, modbus, next) {
 
     let startAt = 8;
     const result = {};
-    for (let i = 0; i < numOfObjects; i++) {
+    // The modbus specification states that numOfObjects is the number of
+    // objects in the response, but the example on page 45 shows the total
+    // number over all responses. Therefore be careful about reading more data than available
+    for (let i = 0; i < numOfObjects && startAt < data.length; i++) {
         const objectId = parseInt(data.readUInt8(startAt));
         const objectLength = parseInt(data.readUInt8(startAt + 1));
         const startOfData = startAt + 2;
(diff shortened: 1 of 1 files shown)


yaacov:
thanks :green_heart:

PR #558, opened 2024-06-16
feat: buf.slice() method has been deprecated

FlyBooks:


yaacov:
Ref: https://nodejs.org/api/buffer.html#bufslicestart-end

yaacov:
Thanks :green_heart:

PR #559, opened 2024-07-24
Make Sockets accessible with TypeScript

cplepage:
Trying to keep track of many sockets are connected to the current `ServerTCP`/`ServerSerial` instance.

While the `socks` property is already accessible, since it's not declared in the `.d.ts` files, TypeScript displays errors when trying to access the property.


yaacov:
Hi, thank you for the pull request
Can you write a description:
Why this is needed
What does it do
How it does it
Example of using it ...
Screenshots if relevant ... ?

PR #562, opened 2024-08-16
feat: Allow close method to be used with Promises or callbacks

grazianogrespan:
First of all a big thank you for this fantastic module!
I was exploring the example folder looking for the `close()` method and I observed that it might be used in a way that suggests synchronous behavior. [polling_TCP.js](https://github.com/yaacov/node-modbus-serial/blob/master/examples/polling_TCP.js#L41)

Checking the implementation of the `close()` method I found out that it uses the Node.JS net.Socket class and in particular the `socket.end()` method which does not guarantee that the socket is fully closed immediately after calling `socket.end()`.

This pull request introduces Promise support to the close method and should not affect the existing usage. 
A callback can still be passed as argument.

### DIFF
--- a/apis/promise.js
+++ b/apis/promise.js
@@ -23,20 +23,28 @@
  * @private
  */
 const _convert = function(f) {
-    const converted = function(address, arg, next) {
+    const converted = function(...args) {
         const client = this;
         const id = this._unitID;
 
-        /* the function check for a callback
-         * if we have a callback, use it
-         * o/w build a promise.
-         */
-        if (next) {
-            // if we have a callback, use the callback
-            f.bind(client)(id, address, arg, next);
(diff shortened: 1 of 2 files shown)


yaacov:
@grazianogrespan hi, thank you for the pull request :green_heart: 

can you move the code to the "promise" file ? see https://github.com/yaacov/node-modbus-serial/blob/master/apis/promise.js#L63

grazianogrespan:
Thanks for your feedback @yaacov !
On the [promise.js](https://github.com/yaacov/node-modbus-serial/blob/master/apis/promise.js) file I tried to use the _convert() method but I had to adjust it to make it work with different argument structures.
Let me know if this is fine for you :pray:

PR #563, opened 2024-08-22
fix: make callback of open, close and destroy optional

kingyue737:


yaacov:
Thank you for the pull request :green_heart:

Issue #540, opened 2024-01-08
Combination of modbus requests in the same TCP Frame.


rvbatista:
Modbus TCP specification determines that only one modbus request needs to be inside a TCP Frame or packet.
Looks like that node-modbus-serial don't use any kind of mechanism to ensure that, correct?
We use node-modbus-serial in FUXA Scada, and one of the users have reported this issue.
https://github.com/frangoteam/FUXA/issues/1030
Off course we can try to handle it on our usage of the library, but looks like that more people can suffer with that issue on their projects.

teddy1565:
```js
/**
     * Write a Modbus "Force Multiple Coils" (FC=15) to serial port.
     *
     * @param {number} address the slave unit address.
     * @param {number} dataAddress the Data Address of the first coil.
     * @param {Array} array the array of boolean states to write to coils.
     * @param {Function} next the function to call next.
     */
    writeFC15(address, dataAddress, array, next) {
        // check port is actually open before attempting write
        if (this.isOpen !== true) {
            if (next) next(new PortNotOpenError());
            return;
        }

        // sanity check
        if (typeof address === "undefined" || typeof dataAddress === "undefined") {
            if (next) next(new BadAddressError());
            return;
        }

        const code = 15;
        let i = 0;

        // set state variables
        this._transactions[this._port._transactionIdWrite] = {
            nextAddress: address,
            nextCode: code,
            nextLength: 8,
            next: next
        };

        const dataBytes = Math.ceil(array.length / 8);
        const codeLength = 7 + dataBytes;
        const buf = Buffer.alloc(codeLength + 2); // add 2 crc bytes

        buf.writeUInt8(address, 0);
        buf.writeUInt8(code, 1);
        buf.writeUInt16BE(dataAddress, 2);
        buf.writeUInt16BE(array.length, 4);
        buf.writeUInt8(dataBytes, 6);

        // clear the data bytes before writing bits data
        for (i = 0; i < dataBytes; i++) {
            buf.writeUInt8(0, 7 + i);
        }

        for (i = 0; i < array.length; i++) {
            // buffer bits are already all zero (0)
            // only set the ones set to one (1)
            if (array[i]) {
                buf.writeBit(1, i, 7);
            }
        }

        // add crc bytes to buffer
        buf.writeUInt16LE(crc16(buf.slice(0, -2)), codeLength);

        // write buffer to serial port
        _writeBufferToPort.call(this, buf, this._port._transactionIdWrite);
    }

[...]

PR #564, opened 2024-09-02
Reduce the chances of multiple Modbus APUs per TCP frame

Abestanis:
This attempts to reduce the likelihood of the problem described in #540, but unfortunately doesn't completely fix it.

* For every data we sent via the Modbus TCP socket we now wait for the [`drain` event](https://nodejs.org/api/net.html#event-drain) of the previous data in the hopes that the new data does not get merged into the same TCP frame.
* We set the [`NoDelay` flag](https://nodejs.org/api/net.html#socketsetnodelaynodelay).

Both changes seem to reduce the amount of TCP frames with more than one Modbus APU, but under very high throughput the can still be some of them. There seems to be no way of guaranteeing separate TCP packages short of using a raw socket and implementing the entire TCP stack manually, so I personally think this is a design flaw of the Modbus protocol and in the end I choose to fix the problem on the Modbus server side by accepting TCP packages with multiple APUs. Also, all other Modbus implementation which I found that were not written for bare metal have exactly the same problem.

I'm posting this here in case you still want to have an improvement over the current state. Sorry about all the whitespace changes, that was the auto formatter of my IDE.

### DIFF
--- a/ports/tcpport.js
+++ b/ports/tcpport.js
@@ -48,7 +48,7 @@ class TcpPort extends EventEmitter {
         /** @type {net.Socket?} - Optional custom socket */
         this._externalSocket = null;
 
-        if(typeof ip === "object") {
+        if (typeof ip === "object") {
             options = ip;
             ip = undefined;
         }
@@ -62,7 +62,7 @@ class TcpPort extends EventEmitter {
         }
 
         /** @type {net.TcpSocketConnectOpts} - Options for net.connect(). */
-        this.connectOptions =  {
+        this.connectOptions = {
             // Default options
             ...{
                 host: ip || options.ip,
@@ -72,8 +72,8 @@ class TcpPort extends EventEmitter {
             ...options
         };
 
-        if(options.socket) {
-            if(options.socket instanceof net.Socket) {
+        if (options.socket) {
+            if (options.socket instanceof net.Socket) {
                 this._externalSocket = options.socket;
                 this.openFlag = this._externalSocket.readyState === "opening" || this._externalSocket.readyState === "open";
             } else {
@@ -92,6 +92,7 @@ class TcpPort extends EventEmitter {
 
         // init a socket
         this._client = this._externalSocket || new net.Socket(this.socketOpts);
+        this._writeCompleted = Promise.resolve();
 
         if (options.timeout) this._client.setTimeout(options.timeout);
 
@@ -174,10 +175,10 @@ class TcpPort extends EventEmitter {
      * @param {(err?: Error) => void} callback
      */
     open(callback) {
-        if(this._externalSocket === null) {
+        if (this._externalSocket === null) {
             this.callback = callback;
             this._client.connect(this.connectOptions);
-        } else if(this.openFlag) {
+        } else if (this.openFlag) {
             modbusSerialDebug("TCP port: external socket is opened");
             callback(); // go ahead to setup existing socket
         } else {
@@ -214,7 +215,7 @@ class TcpPort extends EventEmitter {
      * @param {Buffer} data
      */
     write(data) {
-        if(data.length < MIN_DATA_LENGTH) {
+        if (data.length < MIN_DATA_LENGTH) {
             modbusSerialDebug("expected length of data is to small - minimum is " + MIN_DATA_LENGTH);
             return;
         }
@@ -240,7 +241,15 @@ class TcpPort extends EventEmitter {
         });
 
         // send buffer to slave
-        this._client.write(buffer);
+        this._writeCompleted = new Promise((resolve) => {
+            this._writeCompleted.finally(() => {
+                if (this._client.write(buffer)) {
+                    resolve();
+                } else {
+                    this._client.once("drain", resolve);
+                }
+            });
+        });
 
         // set next transaction id
         this._transactionIdWrite = (this._transactionIdWrite + 1) % MAX_TRANSACTIONS;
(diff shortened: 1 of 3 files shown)


ports/tcpport.js
```diff
@@ -240,7 +242,15 @@ class TcpPort extends EventEmitter {
...
-        this._client.write(buffer);
+        this._writeCompleted = new Promise((resolve, reject) => {
+            this._writeCompleted.finally(() => {
+                if (this._client.write(buffer)) {
+                    resolve();
+                } else {
+                    this._client.once('drain', resolve);
+                }
+            });
+        });
```
yaacov, line 253:
```suggestion
       this._writeCompleted = new Promise((resolve, reject) => {
            if (this._client.write(buffer)) {
                resolve();
            } else {
                this._client.once('drain', resolve);
            }
        });

        // Wait for the write operation to complete
        try {
            await this._writeCompleted;
        } catch (error) {
           modbusSerialDebug('An error occurred during the write operation:', error);
            throw error;
        }
```

yaacov:
hi, thank you for the pull request :green_heart: 
I added a suggestion, can you take a look ?

ports/tcpport.js
Abestanis: Unfortunately, `write` is not `async`. so we can't await in it. It is my understanding that `_client.write` will emit an `error` event on error instead of throwing, although I can't find any documentation about that online.

Abestanis: I think there might be a misconception here:

The `this._writeCompleted.finally` is not waiting on itself, it's waiting on the previous promise that was stored in `this._writeCompleted` and then overwrites `this._writeCompleted` with the new promise. We need to have a Promise because we need a way to signal that we are done from the callback passed to write, which can be called asynchronously.

The only thing I can think of is do the assignment on another line, do you think this would help?
```js
let newWritePromise = this._writeCompleted = new Promise((resolve) => {
    this._writeCompleted.finally(/* ... */);
};
this._writeCompleted = newWritePromise;
```

yaacov: i have some concerns with the new suggestion ( valid to current suggestion too )

`this._writeCompleted` is overwritten each time a new write is initiated. This means if multiple writes are attempted quickly in succession, each write would create a new promise, but the old promise (from a previous write) could be lost. If the older promise hasn't resolved yet (due to waiting for the drain event), it could lead to race conditions where multiple writes might overlap.

If there’s an error in the write process (such as a disconnection), the promise will never resolve or reject, potentially leading to the system hanging indefinitely. For example, if you're managing network connections, where interruptions are common.

can you address the concerns, and since this flow is not clear, can you add comments helping to understand the flow, e.g. what is the current promise we handle, when and why we replace it etc...

Abestanis: > but the old promise (from a previous write) could be lost.

I think this is not possible. The function passed as an argument to the `Promise` constructor is executed synchronously before the constructor returns. And as far as I understand Javascript is single threaded, so there is no chance of a race condition where two `TcpPort.write` calls happen simultaneously.

> If the older promise hasn't resolved yet (due to waiting for the drain event), it could lead to race conditions where multiple writes might overlap.

This should also not be possible, because we only call `Socket.write` in the `finally` call of the previous Promise, so there should be no overlap.

> If there’s an error in the write process (such as a disconnection), the promise will never resolve or reject

It is my understanding that `Socket.write` will never throw, but just trigger an `error` event and then return. But I will add a `try catch` to make sure the promise is resolved.

> can you add comments helping to understand the flow

Will do.

ports/tcpport.js
```diff
@@ -240,7 +242,26 @@ class TcpPort extends EventEmitter {
...
+            // Wait for the completion of any write that happened before.
+            previousWritePromise.finally(() => {
+                try {
+                    // The previous write succeeded, write the new buffer.
+                    if (this._client.write(buffer)) {
+                        // Mark this write as complete.
+                        resolveNewWrite();
+                    } else {
+                        // Wait for one `drain` event to mark this write as complete.
+                        this._client.once("drain", resolveNewWrite);
```
yaacov, line 256: do we need to address a case where "drain" is not called ? for example connection fail ?

Abestanis: I don't think that's necessary. The [`drain` event documentation](https://nodejs.org/api/stream.html#event-drain) says

> If a call to [stream.write(chunk)](https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback) returns false, the 'drain' event will be emitted when it is appropriate to resume writing data to the stream.

I think this means that we are guaranteed to get a `drain` event, if `Socket.write` returns `false`. You can also see that the usage example in the documentation does not include any specific consideration for the `error` event.

But in case a user of the `TcpPort` decides to close and then re-open the the port, I overwrote the promise in the `connect` event so that we are sure to not be blocked by writes on a closed port.

yaacov:
LGTM, can you make the tests not fail ?

Abestanis:
> LGTM, can you make the tests not fail ?

The tests seem to assume that a call to `write` will set the `_buffer` directly, but because write is called in a the `finally` handler, this is no longer the case.

I'm not too familiar with javascript testing, could you please give me some pointers on how to approach this?

Abestanis:
Any hints on how to fix the tests would be appreciated.

yaacov:
> Any hints on how to fix the tests would be appreciated.

When running locally, does the error msgs makes sense to you ?
Did you identify the failing test on your local dev environment ?

Abestanis:
The error message did make sense to me, in all cases it's because the tests were expecting the write to have an immediate effect and not waiting for the new promise. Once I realized that I could just attach and wait for the private internal write promise in the text it became clear how to fix them. They pass locally now, let's hope CI is also happy.

PR #576, opened 2025-02-05
fix: catch error thrown in getMultipleInputRegisters

danieljpwtg:
Similar fix as in this PR https://github.com/yaacov/node-modbus-serial/pull/543 just for getMultipleInputRegisters


yaacov:
field again, don't wary, try to run `npm run test`

Issue #581, opened 2025-02-21
Compilation error with typescript


AuspeXeu:
![Image](https://github.com/user-attachments/assets/288fa5e1-7273-46ea-9949-425062331ed4)

yaacov:
thank you for the issue, can you make a pull request fixing this issue ?

tonyho:
Same issue

PR #585, opened 2025-03-04
Update ServerSerial.d.ts

AuspeXeu:
Remaining 50% of #581


yaacov:
> Remaining 50% 

:-) Nice !

PR #587, opened 2025-03-17
fix: added close event for rtu buffered port

Zelan89:
It seems like in rtubufferedport.js the "close" event was missed, so the 
client.on('close', () => {
    console.log('Modbus closed');
});
never fires up.
Changed it according to tcprtubufferedport.js


yaacov:
Thanks :green_heart:

Zelan89:
Thank you for the quick reply.
Is it possible to release new npm package version?

PR #588, opened 2025-03-19
Update ModbusRTU.d.ts

chowchin:
Adding timeout and socket in the TelnetPortOptions


yaacov:
Thank you :green_heart:

PR #613, opened 2026-06-10
feat: add register32bit option for devices with native 32-bit registers

stuartm-hm:
## Summary

- Adds a `register32bit` connection option that treats **every** register on the device as a single 32-bit (4-byte) value, rather than the standard Modbus 16-bit width
- This is distinct from the existing Enron address-range approach — `register32bit: true` applies to all addresses unconditionally, with no table configuration required
- All affected paths (FC3/4 response parsing, FC6 response parsing, read-request expected-length calculation, FC6 write framing, FC16 write framing) are covered

## What changed

| File | Change |
|---|---|
| `apis/connection.js` | Read `options.register32bit` and store as `this._register32bit` in `connectRTU`, `connectTCP`, and `connectRTUBuffered` |
| `index.js` | Prepend `modbus._register32bit \|\|` (or `this._register32bit \|\|`) to all six Enron-gated 32-bit code paths |
| `ModbusRTU.d.ts` | Add `register32bit?: boolean` to `SerialPortOptions` and `TcpPortOptions` |

## Usage

```js
client.connectRTU("COM1", { baudRate: 9600, register32bit: true }, callback);
// or TCP:
client.connectTCP("192.168.1.10", { port: 502, register32bit: true }, callback);

// Read one register → returns a single UInt32, not two UInt16s
client.readHoldingRegisters(100, 1, (err, data) => {
    console.log(data.data[0]); // e.g. 305419896 (0x12345678)
});

// Write one 32-bit register
client.writeFC6(1, 100, 0x12345678, callback);

// Write multiple 32-bit registers
client.writeFC16(1, 100, [0x12345678, 0xDEADBEEF], callback);
```

## Notes for reviewer

- The existing Enron behaviour is completely unchanged; `register32bit` is checked first as a short-circuit
- When passing a raw `Buffer` to `writeFC16`, callers are expected to pre-pack 4 bytes per register (the byte-length is divided by 4 instead of 2)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added 32-bit register support for Modbus RTU and TCP connections via new optional configuration parameter
  * Enhanced register parsing and write operations to handle both 16-bit and 32-bit register values


stuartm-hm:
Closing — opened by mistake, not ready for upstream.

Issue #599, opened 2025-12-30
Modbus Exception 3


HenriqueAurelio:
# Intermittent Modbus Exception 3 in Production - Critical Coil Operations Lost

## Description

I'm experiencing intermittent **Modbus Exception 3 (Illegal data value)** errors when reading from specific PLCs in production environments. This issue is critical because **missing coil reads causes my application to lose core functionality** (gate control actions are not executed).

The same application code works reliably with most clients (15+ PLCs), but with some specific clients, Exception 3 errors occur frequently, disrupting operations.

## Environment

- **Library Version:** `modbus-serial@8.0.26`
- **Node.js Version:** `v18`
- **Protocol:** Modbus TCP
- **Runtime:** Node.js in Docker containers
- **Network:** Production environments with varying network quality
- **Scale:** 20+ PLCs running the same codebase

### Error Message
```
Error: Modbus exception 3: Illegal data value (value cannot be written to this register)
    at ModbusRTU._onReceive (/app/node_modules/modbus-serial/index.js:474:21)
    at TcpPort.emit (node:events:517:28)
    at TcpPort.emit (node:domain:489:12)
    at Socket.<anonymous> (/app/node_modules/modbus-serial/ports/tcpport.js:123:22)
    ...
```

### Frequency & Pattern
- ❌ **Occurs frequently on ~5 specific client PLCs** (approximately 25% of deployments)
- ✅ **Rarely/never occurs on 15+ other client PLCs** using identical code
- ❌ **Most common when reading coils**, but also happens with discrete inputs and holding registers
- ❌ **Intermittent** - not every request fails, making it hard to reproduce consistently
- ⚠️ **More frequent during peak hours** or suspected network congestion periods

### Critical Impact
**I cannot afford to lose coil operations** - when a coil read fails with Exception 3, my application misses gate control actions, which is the core functionality of the system. This results in:
- Gates not opening when vehicles arrive
- Operational delays and safety concerns
- Lost revenue and customer complaints

## Code Sample

```javascript
const ModbusRTU = require("modbus-serial");
const client = new ModbusRTU();

// Connection setup
client.setTimeout(5000); // Increased from default 1000ms
await client.connectTCP(plcHost, { port: 502 });
client.setID(1);

// Polling loop (runs every 1 second)
setInterval(async () => {
  try {
    // Read discrete inputs
    const inputs = await client.readDiscreteInputs(0, 7);
[...]

PR #623, opened 2026-08-21
Validate the MBAP header before trusting its declared length

Poseidonas:
Follow-up to #599, where an intermittent `Modbus exception 3` turned out to sit on top of a parser problem.

## What happens today

`ports/tcpport.js` reads the MBAP length field and acts on it without checking whether it is possible:

```js
length = self._clientRcvData.readUInt16BE(4);
if (self._clientRcvData.length < (length + MIN_MBAP_LENGTH)) {
    return;
}
```

Modbus RTU protects the frame with a CRC on the wire. Modbus TCP does not — it relies on TCP, and this port then computes the CRC over the bytes it has just received and writes it into the buffer itself, so the later check in `index.js` matches by construction whatever was corrupted. The result is that a damaged header is indistinguishable from a valid one.

When that happens the connection does not degrade, it stops. The guard returns early on every subsequent read, valid responses keep appending to `_clientRcvData`, nothing is ever emitted again, and every transaction times out until the process is restarted. There is no upper bound on that buffer either.

Measured against a mock device that answers normally, with bytes corrupted only in transit:

```
                          request 1   2         3         4
baseline, no corruption   OK          OK        OK        OK
MBAP length = 71          Timed out   Timed out Timed out Timed out
MBAP length = 65535       Timed out   Timed out Timed out Timed out
MBAP length = 0           Timed out   Timed out Timed out Timed out
```

## What this changes

The spec puts the MBAP length at Unit ID (1) + PDU (1..253), so anything outside 2..254 cannot come from a conforming device, and the protocol identifier is always 0. Both are now checked before the length is trusted, and the buffer is discarded so the stream can resynchronise on the next response.

A length that is in range but wrong — 71 in the reported case — cannot be recognised from the header alone. It only reveals itself by never completing, so a frame still pending after one second is dropped as well. A conforming response is at most 260 bytes and arrives within milliseconds of its first byte, so this does not fire on healthy traffic, including slow links.

Same scenarios after the change:

```
                          request 1   2         3         4
MBAP length = 71          Timed out   Timed out Timed out OK  (recovers at 1.6s)
MBAP length = 65535       Timed out   OK        OK        OK
MBAP length = 0           Timed out   OK        OK        OK
baseline, no corruption   OK          OK        OK        OK
```

The damaged frame is still lost — nothing can recover it — but the connection survives it.

## Tests

Five cases added to `test/ports/tcpport.test.js`: length above the maximum, below the minimum, a non-zero protocol identifier, the receive buffer not growing without bound, and an ordinary response still parsing. The existing suite passes unchanged (172 total).

## Notes for review

Both discards are silent, matching how the port already handles data it cannot use. If you would rather surface them as an `error` event, or make the partial-frame window configurable alongside the existing timeout, both are small changes — I did not want to assume which fits the library's conventions.

I have also left the forged-exception case out of this PR. A corrupted frame can still present as a Modbus exception the device never sent, and a genuine exception response is always exactly three bytes, so a length mismatch there would be detectable. It felt like a separate change and a separate discussion.


Poseidonas:
Closing — I targeted the wrong base branch, so the diff picked up unrelated history. Reopening against `main` with only the two files this touches. Sorry for the noise.

PR #624, opened 2026-08-21
Validate the MBAP header before trusting its declared length

Poseidonas:
Follow-up to #599, where an intermittent `Modbus exception 3` turned out to sit on top of a parser problem.

## What happens today

`ports/tcpport.js` reads the MBAP length field and acts on it without checking whether it is possible:

```js
length = self._clientRcvData.readUInt16BE(4);
if (self._clientRcvData.length < (length + MIN_MBAP_LENGTH)) {
    return;
}
```

Modbus RTU protects the frame with a CRC on the wire. Modbus TCP does not — it relies on TCP, and this port then computes the CRC over the bytes it has just received and writes it into the buffer itself, so the later check in `index.js` matches by construction whatever was corrupted. The result is that a damaged header is indistinguishable from a valid one.

When that happens the connection does not degrade, it stops. The guard returns early on every subsequent read, valid responses keep appending to `_clientRcvData`, nothing is ever emitted again, and every transaction times out until the process is restarted. There is no upper bound on that buffer either.

Measured against a mock device that answers normally, with bytes corrupted only in transit:

```
                          request 1   2         3         4
baseline, no corruption   OK          OK        OK        OK
MBAP length = 71          Timed out   Timed out Timed out Timed out
MBAP length = 65535       Timed out   Timed out Timed out Timed out
MBAP length = 0           Timed out   Timed out Timed out Timed out
```

## What this changes

The spec puts the MBAP length at Unit ID (1) + PDU (1..253), so anything outside 2..254 cannot come from a conforming device, and the protocol identifier is always 0. Both are now checked before the length is trusted, and the buffer is discarded so the stream can resynchronise on the next response.

A length that is in range but wrong — 71 in the reported case — cannot be recognised from the header alone. It only reveals itself by never completing, so a frame still pending after one second is dropped as well. A conforming response is at most 260 bytes and arrives within milliseconds of its first byte, so this does not fire on healthy traffic, including slow links.

Same scenarios after the change:

```
                          request 1   2         3         4
MBAP length = 71          Timed out   Timed out Timed out OK  (recovers at 1.6s)
MBAP length = 65535       Timed out   OK        OK        OK
MBAP length = 0           Timed out   OK        OK        OK
baseline, no corruption   OK          OK        OK        OK
```

The damaged frame is still lost — nothing can recover it — but the connection survives it.

## Tests

Five cases added to `test/ports/tcpport.test.js`: length above the maximum, below the minimum, a non-zero protocol identifier, the receive buffer not growing without bound, and an ordinary response still parsing. The existing suite passes unchanged (172 total).

## Notes for review

Both discards are silent, matching how the port already handles data it cannot use. If you would rather surface them as an `error` event, or make the partial-frame window configurable alongside the existing timeout, both are small changes — I did not want to assume which fits the library's conventions.

I have also left the forged-exception case out of this PR. A corrupted frame can still present as a Modbus exception the device never sent, and a genuine exception response is always exactly three bytes, so a length mismatch there would be detectable. It felt like a separate change and a separate discussion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TCP communication when incomplete or corrupted frames are received.
  * Invalid protocol headers and out-of-range frame lengths are now safely skipped while valid data is recovered.
  * Partial frames are cleared after a timeout to prevent stale data from blocking communication.
  * Valid responses continue to be parsed correctly after corrupted or incomplete data.
  * Buffered data is cleaned up when connections close or are destroyed.
  * Improved recovery when corrupted data is followed by a complete frame or an exact header boundary.


Poseidonas:
Both findings were right, and I reproduced each one before changing anything rather than taking them on trust.

**Discarding recoverable bytes.** Confirmed: a corrupted frame and a valid response delivered in a single `receive()` call emitted nothing at all, so the good response was lost along with the bad one.

The parser now searches for the next offset that begins a frame the buffer can actually satisfy and keeps everything from there. One detail worth flagging, because my first attempt got it wrong: scanning only for a plausible protocol identifier and length is not sufficient. Random payload bytes match that test often — my own test for `length = 0` started failing because the scan locked onto a false header inside the damaged frame and carried the misalignment forward. Requiring the declared frame to end within the bytes already held is a much stronger signal, and that is what is implemented.

**Partial frames never expiring.** Confirmed on both counts. With a peer that sent a partial frame and then went quiet, the buffer still held its 8 bytes after 1.5 seconds, because the deadline was only consulted when another `data` event arrived. And the loop bound was `>` rather than `>=`, so a buffer holding exactly the six header bytes was never examined — you were right that the branch was unreachable in that case.

It is now a real timer, armed when an incomplete but valid header is buffered, cancelled when the frame completes, when the header proves invalid, and on `close()` and `destroy()`. It is `unref`'d so a damaged frame cannot hold the event loop open. On expiry it resynchronises rather than clearing, for the same reason as above.

Three regression tests added, matching what you asked for:

```
✔ should keep a valid response delivered alongside a corrupted one
✔ should release a partial frame when the peer goes silent
✔ should examine a header delivered as exactly six bytes
```

175 tests pass. End-to-end against a mock device, before and after:

```
                          request 1   2         3         4
before, length = 71       Timed out   Timed out Timed out Timed out   (never recovers)
after,  length = 71       Timed out   Timed out Timed out OK
after,  length = 65535    Timed out   OK        OK        OK
after,  length = 0        Timed out   OK        OK        OK
after,  no corruption     OK          OK        OK        OK
```

Thank you for the review — the first version would have traded one failure mode for a quieter one.

Poseidonas:
Both correct again, and both reproduced before I touched anything.

**Response queued behind a stale frame.** Confirmed: an in-range but wrong length followed by a complete valid response in one read left the good response buffered indefinitely — after 1.4 seconds, zero frames emitted and 12 bytes still held. The timer resynchronised but nothing parsed the buffer afterwards.

The receive loop is now `_parseReceivedData()`, called from the `data` handler and again by the timer once it has realigned. Same case now emits the response.

**Timer surviving a remote close.** Confirmed: the timer was still armed after the socket `close` handler ran. Cancelled there too, alongside `close()` and `destroy()`.

Two regression tests added:

```
✔ should deliver a response queued behind a stale partial frame
✔ should cancel the partial frame timer when the peer closes
```

177 tests pass, lint clean. One caveat on the second test: driving `_client.emit("close")` directly re-entered the `open` callback through the mock and tripped mocha's `done()` guard, so the test exercises `close()` rather than a peer-initiated close. The code path is cancelled in the socket handler and I verified that separately outside the suite, but the test covers the weaker of the two. If you would prefer it done properly, extending the net mock to simulate a remote disconnect would be the way, and I am happy to do that if it is worth the addition.

End-to-end, the recovery is now a request earlier than the previous round:

```
                          request 1   2         3         4
before any fix, len = 71  Timed out   Timed out Timed out Timed out
now,            len = 71  Timed out   Timed out OK        OK
now,            len = 65535 Timed out OK        OK        OK
now,            len = 0     Timed out OK        OK        OK
now,            no corruption  OK     OK        OK        OK
```

Poseidonas:
This one I am answering rather than changing, because I measured the alternative and it is worse. Happy to be overruled.

The concern is correct: if TCP splits a genuine response across reads and corruption precedes it, no complete frame is visible yet, `findNextHeader` returns -1, and those bytes are dropped. I reproduced it — a corrupted frame plus the first six bytes of a valid response in one read, the remaining six in the next, and the valid response is lost.

The obvious fix is to fall back to the first merely plausible header when no complete frame is found. I implemented that and it broke an existing test, which turned out to be the useful part:

```
case                          strict   loose
corrupt garbage on its own    -1       offset 1  ← declares 17 bytes, only 11 present
corrupt + response in pieces  -1       offset 12 ← correct
```

On a 12-byte corrupted frame the loose search locks onto offset 1, because two arbitrary payload bytes read as protocol 0 and a length inside 2..254. It then resynchronises onto payload and the next read is parsed against that misalignment. A plausible header is a weak signal — random bytes satisfy it frequently — so acting on one carries the misalignment forward instead of clearing it.

That matters more than it looks, because a fabricated frame is exactly the failure this PR exists to prevent. In #599 the reported symptom was `Modbus exception 3` from a device that never raised one, produced by corrupted bytes being parsed as a valid response. A loose resynchronisation manufactures the same class of frame deliberately.

So the trade is: strict loses one recoverable response to a timeout, loose risks emitting a frame that was never sent. A timeout is visible, retryable and honest; a fabricated exception is silent and misleading. I have kept strict and documented the cost in the function comment so the next reader does not have to rediscover it.

If you would rather have the recovery, two options that avoid the guessing:

- Fall back to a plausible header **only** when it is the last one in the buffer and every remaining byte belongs to the frame it declares. Narrower, though still a guess.
- Have the port track outstanding transaction identifiers and accept an offset only when the two bytes there match a request still in flight. That is a genuine anchor rather than a heuristic, but it means the port knowing about transactions, which is currently `index.js`'s concern — a bigger change than I would make uninvited.

Either is straightforward if you prefer one; I did not want to pick an architecture on your behalf.

Issue #573, opened 2025-01-20
Data length error


harshit-malik7:
{"level":"info","message":"Timestamp: 2025-01-17T22:02:20.880Z, msg: \"HVAC Id: 1 Fetching HVAC mode\", correlationId: 9432, function: getHVACMode, level: info"}
{"level":"error","message":"Timestamp: 2025-01-17T22:02:20.962Z, msg: \"HVAC Id: 1 Failed to get HVAC mode because of Error: Data length error, expected 7 got 6\", correlationId: 9432, function: getHVACMode, level: error"}
const response = await requestedHVAC.readHoldingRegisters(803,1);
            HVACMode = response.data[0];
it seems to happen sometimes but sometimes it goes through. Same issue with readcoils and readDiscreteInputs as well.

Poseidonas:
I think I can account for this one. It is not the device sending a short reply — it is a reply from an *earlier* request arriving after that request had already timed out.

**Reproduced**

```
1) readCoils on txn=1, expects 6 bytes  → timed out, device answers late
2) transaction id wraps back to 1 after 256 requests
3) readHoldingRegisters on txn=1, expects 7 bytes
   the late readCoils reply arrives and is matched to it
   → Data length error, expected 7 got 6
```

**Why the two get confused**

`_transactionIdWrite` wraps at `MAX_TRANSACTIONS` (256) in `ports/tcpport.js:429`, and entries in `this._transactions` are never removed — the object is only ever reset wholesale on reconnect. So after 256 requests an identifier comes round again, and a reply still in flight from the previous holder of that identifier looks, to `_onReceive`, exactly like the reply to the current one. The transaction id is the only thing distinguishing them and it is no longer unique in time.

That fits your symptoms: intermittent rather than constant, since it needs a slow reply and a wrap to coincide; and appearing across `readCoils`, `readDiscreteInputs` and `readHoldingRegisters` alike, since any pairing of two different expected lengths produces it. With a busy HVAC controller and correlation ids in your logs, 256 requests come round quickly.

**Why the message is misleading**

In `index.js` the length check sits at line 544 and the function code check at line 567. The stale reply fails the length check first, so you are told about a byte count when the real problem is that the frame belongs to a different request. Moving the function code check ahead of the length check reports what actually happened:

```
before:  Data length error, expected 7 got 6
after:   Unexpected data error, expected code 3 got 1
```

[...]

PR #625, opened 2026-08-21
Report a function code mismatch before a length mismatch

Poseidonas:
Addresses the misleading message in #573, where the reproduction and the underlying cause are written up.

## What happens

A reply that arrives after its request has timed out is matched to whichever request now holds the same transaction identifier. Identifiers wrap at `MAX_TRANSACTIONS` (256) and entries in `_transactions` are never removed, so an identifier stops being unique in time and two requests become indistinguishable to `_onReceive`.

Reproduced with a device that answers late:

```
1) readCoils on txn=1, expects 6 bytes → times out, device answers late
2) identifier wraps back to 1 after 256 requests
3) readHoldingRegisters on txn=1, expects 7 bytes
   the late readCoils reply arrives and is matched to it
   → Data length error, expected 7 got 6
```

## What this changes

Only the message. The length check sat ahead of the function code check, so a frame belonging to another request failed on its byte count first, and the caller was pointed at the wrong thing:

```
before   Data length error, expected 7 got 6
after    Unexpected data error, expected code 3 got 1
```

Moving the function code check first says what actually happened. The reordering is two blocks swapped; no logic is altered.

## Tests

`testport` gains unit 7, answering with a different function code and a length to match, which stands in for a stale reply. One test asserts the code mismatch is reported and the length error is not. Existing suite passes (168 tests, lint clean).

## What this deliberately does not do

It does not stop the mismatch. Preventing it means changing how identifiers are allocated — removing a transaction once it settles and holding its identifier back until the socket has been quiet, or refusing to reuse one whose transaction is still outstanding. Both are design decisions about the port's contract rather than a fix a passer-by should pick, so I have left them alone and described the options in #573.

If you would rather have the prevention as well, or would rather this waited for it, I am happy either way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response validation to identify unexpected function codes before checking response length.
  * Corrected error messages for responses with mismatched function codes, preventing misleading validation errors.

* **Tests**
  * Added coverage for responses that contain both an unexpected function code and an invalid payload length.
  * Added response simulation to verify accurate handling of malformed function-code responses.


Poseidonas:
Reordering two checks can quietly swallow a case that used to be reported correctly, so I walked every error path in `testport` before and after. Behaviour is unchanged except for the one this targets:

| case | unit | result |
| --- | --- | --- |
| correct reply | 1 | passes |
| genuinely short reply | 2 | `Data length error, expected 11 got 6` — unchanged |
| bad CRC | 3 | `CRC error` — unchanged |
| wrong unit number | 4 | `Unexpected data error, expected address…` — unchanged |
| device exception | 5 | `Modbus exception 4` — unchanged |
| no reply | 6 | `Timed out` — unchanged |
| foreign function code | 7 | `Unexpected data error, expected code 3 got 1` — **new** |

The one worth checking is unit 2: a reply that really is short still reports the length error, so the length check has not been shadowed. It only yields when the function code already says the frame belongs to a different request.

On bounds: the minimum-length guard at `index.js:486` still runs first, so `data.readUInt8(1)` at line 505 cannot read past the end. The reordering happens well after both.

Poseidonas:
Both fixed.

The comment said the reply had "a length to match" while the code drops a byte — it now says one byte short, which is what unit 7 actually does.

The assertion was only proving the absence of a length error, so it would have passed on any other unexpected-data message. It now asserts the exact string, including both function codes:

```
Unexpected data error, expected code 3 got 1
```

Lint clean, 168 tests passing.

